PR #4, opened 2020-07-14
Modifies jvm system property to look for graalvm

eginez:
Some version of the GraalVM JDK might not have `java.vendor.version` set

### DIFF
--- a/src/main/java/io/micronaut/gradle/graalvm/GraalUtil.java
+++ b/src/main/java/io/micronaut/gradle/graalvm/GraalUtil.java
@@ -11,7 +11,7 @@ public final class GraalUtil {
      * @return Return whether the JVM in use a GraalVM JVM.
      */
     public static boolean isGraalJVM() {
-        String vv = System.getProperty("java.vendor.version");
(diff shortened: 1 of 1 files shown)


graemerocher:
Thanks @eginez 

Can you sign the CLA? Oracle has company wide permission to do so

graemerocher:
hmm I think we need both checks. `jvmci.Compiler` returns `null` for Graal CE

Steinliiippp:
Is it not possible to use the Graal CE to build native images? The ouput of java -verison is:
openjdk version "11.0.8" 2020-07-14
OpenJDK Runtime Environment GraalVM CE 20.2.0 (build 11.0.8+10-jvmci-20.2-b03)
OpenJDK 64-Bit Server VM GraalVM CE 20.2.0 (build 11.0.8+10-jvmci-20.2-b03, mixed mode, sharing)

The output of System.getProperty("jvmci.Compiler") is null and the output of System.getProperty("java.vendor.version") is GraalVM CE 20.2.0. As i saw in the plugin code both outputs are checked. So on my setup the native image task failes.

graemerocher:
GraalVM CE works fine yes

PR #7, opened 2020-08-02
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


sdelamo:
@alvarosanchez not sure if our build plugin already does this.

alvarosanchez:
> @alvarosanchez not sure if our build plugin already does this.

So far, we use both Dependabot and our dependency upgrade mechanism. We haven't decided to retire Dependabot.

PR #18, opened 2020-09-04
Update Open API groupId

ilopmar:
The related Open API test fails at this moment because we need a Micronaut version that includes the new groupId. I've put 2.1.0.M1 in the test (we may want to chage it to 2.1.0.RC1) so once the first 2.1 version is published we can merge this.


ilopmar:
There won't be 2.1 RC version so I'll update this PR with the latest Openapi 2.x version a few days before Micronaut 2.1.0 release.

ilopmar:
This is not necessary anymore.

PR #30, opened 2020-09-12
Make AnnotationProcessingConfigBuilder public

donbeave:
This will fix the warning in IntelliJ IDEA:

Type AnnotationProcessingConfigBuilder! is inaccessible in this context due to: public/*package*/ interface AnnotationProcessingConfigBuilder defined in io.micronaut.gradle in file AnnotationProcessingConfigBuilder.class

image


graemerocher:
Thanks for the contribution!

PR #52, opened 2020-10-03
Use Groovy compiler options from Gradle Api

katoquro:
This patch uses Gradle Api for Groovy Compiler(since Gradle 6) to setup compilation with parameters.


graemerocher:
Thanks for the contribution!

PR #62, opened 2020-10-20
Improve `yum install` command

alvarosanchez:
- Ensures latest versions are installed.
- Removes non-existing or already installed packages.
- Keeps layer as small as possible.


alvarosanchez:
Closing this. There are more improvements to come, but it is better to wait until the Maven and Gradle plugins have a shared codebase.

PR #66, opened 2020-10-26
Allow running tests with the native image server

graemerocher:
This PR starts the native image server and configures Micronaut Test to use so that integration tests can also test the native image build


alvarosanchez:
Interesting approach. Why can't you use `ProcessBuilder#inheritIO()`? To do it on a different thread?

graemerocher:
I did try `ProcessBuilder#inheritIO()` but it didn't seem to work, there was no output from Gradle. I think gradle swaps out the system out some point later in the lifecycle

alvarosanchez (approved):

PR #71, opened 2020-10-29
Add memory options and remove library path from AWS custom runtime

ilopmar:
Adding the library path shouldn't be necessary anymore because with the most recent versions of GraalVM it's not necessary to pass libraries like we used to do with `libsunec.so` in the past.

I've replaced that with the memory settings that we had before in Starter: https://github.com/micronaut-projects/micronaut-starter/pull/379/commits/5a38c0e2ee081a1c48f8c5555b3996eb2e43b8ed


graemerocher:
I don't think hard coding the memory settings is the correct thing to do. The template in launch should be modified to include this setting.

ilopmar:
Is there a way I can pass that value to the plugin?

graemerocher:
Yes, see the docs at https://github.com/micronaut-projects/micronaut-gradle-plugin/

```
dockerfile {
 args("-Xmx128m")
}
dockerfileNative {
 args("-Xmx64m")
}
```

PR #119, opened 2021-01-23
Feature/allow additional docker commands + fix

osh-onstructive:
This fixes the issues found with the original implementation, where the DockerBuildStrategy wasn't taken into account correctly & also the args.


graemerocher:
Thanks for the contribution!

PR #123, opened 2021-01-27
Use alpine-glibc:3.12 as base image

ilopmar:
The `latest` version of `frolvlad/alpine-glibc` (3.13) is not compatible with GraalVM and application fail with a segmentation fault. This PR sets the version to the previous one that works (3.12)

More info: https://github.com/oracle/graal/issues/3159


graemerocher:
PR is failing with manifest not found

ilopmar:
Sorry, I didn't put the right docker image.

PR #125, opened 2021-02-01
Found that the logic was wrong for the check if there was a dockerfile or not.

osh-onstructive:
These I missed during the first implementation, but now tested ... sorry for the mistake.

PR #126, opened 2021-02-02
Build "mostly" static native image and use distroless docker image

ilopmar:
This PR enables the generation of [mostly static native images](https://www.graalvm.org/reference-manual/native-image/StaticImages/#build-a-mostly-static-native-image) and switch from an alpine-glibc docker image to a distroless one.

I've tested in CI and everything is working. I've also tested this deploying to AWS Lambda and it also works.

### DIFF
--- a/src/main/java/io/micronaut/gradle/graalvm/NativeImageTask.java
+++ b/src/main/java/io/micronaut/gradle/graalvm/NativeImageTask.java
@@ -105,6 +105,9 @@ public class NativeImageTask extends AbstractExecTask<NativeImageTask>
         String imageName = getImageName().get();
         args("-H:Name=" + imageName);
 
+        // to build a "mostly" static native image
+        args("-H:+StaticExecutableWithDynamicLibC");
+
         // Adds boolean flags to the command line
         booleanCmds.forEach((property, cmd) -> {
             if (property.getAsBoolean()) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/micronaut/gradle/graalvm/NativeImageTask.java
```diff
@@ -105,6 +105,9 @@ public void configure() {
         String imageName = getImageName().get();
         args("-H:Name=" + imageName);
 
+        // to build a "mostly" static native image
+        args("-H:+StaticExecutableWithDynamicLibC");
```
graemerocher, line 109: I think this argument should only be added automatically if the image is `distroless` since we can't assume this flag for all images

graemerocher (changes requested):

src/main/java/io/micronaut/gradle/graalvm/NativeImageTask.java
ilopmar: Yes, you are right. I found an issue with previous changes, see https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/125#issuecomment-771758146. I'll wait for the user to fix his stuff and then I'll do my changes on top of his work.

ilopmar:
Superseeded by #134

PR #127, opened 2021-02-03
Rework the implementation to allow additional commands to be supplied.

osh-onstructive:
Found that my original implementation was adding too many workarounds, and decided to rework it using primarily the original implementation, but with a different entry point and evaluation time.


graemerocher:
@ilopmar could you review this change solves the recent issues you have found?

ilopmar:
Yes, I'll do it probably tomorrow morning.

ilopmar:
With this PR now the `baseImage("foo")` defined in either `dockerfileNative { }` or `dockerfile { }` is used in the generated dockerfiles, so everything good :)

I've found another issue that was introduced in #78. As it is not related to this, I'll open a separate issue.

PR #140, opened 2021-02-12
Build GraalVM dynamically-linked native images by default

### LINKED ISSUE #139: Build dynamically-linked native images by default
As agreed in the last engineering meeting we should generate by default dynamically-linked native images because it is the default in GraalVM.
We should also document how to generate mostly static native images if the users want that approach because we know they work.

ilopmar:
Fixes #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,26 @@ nativeImage {
 
 **IMPORTANT:** If you update an existing Micronaut application that contains the file `src/main/resources/META-INF/native-image/xxxxx/native-image.properties`, please make sure to delete the properties `-H:Name` and `-H:Class` from the file because they are managed automatically by the plugin.
 
+#### Build "mostly static" native images
+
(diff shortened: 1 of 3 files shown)


graemerocher:
I'm confused I thought we decided to build statically linked images by default?

ilopmar:
We discussed this yesterday (just after you left) and we agreed that by default we should provide dynamically-linked images because that's the default in GraalVM. This started because I'm working on bringing some of the existing functionality in the Gradle plugin to the Maven one. When discussing a few options with @alvarosanchez he though that we shouldn't do it. So we discussed at the engineering meeting.

@alvarosanchez feel free to add your thoughts here.

ilopmar:
@graemerocher more context from the engineering meeting notes: https://github.com/micronaut-projects/micronaut-core/wiki/20210211-Engineering-Meeting#iv%C3%A1n

PR #178, opened 2021-04-09
refactor: delete print statement

sdelamo:
When you execute a gradle task, this is output:

```
 ./gradlew buildNativeLambda

> Configure project :
DOESN'T EXIST!! = native-image
DOESN'T EXIST!! = native-image
```

I don't think we should have this print statement. Or if we want it it should use [Gradle logger](https://docs.gradle.org/current/userguide/logging.html) and it should not appear by default.


parker-mar:
Thanks for the reversion! It was obfuscating our outputs a bit.

PR #179, opened 2021-04-09
Fix of up-to-date check for dockerBuild task and general improvements 

ivyazmitinov:
This merge request fixes issue with up-to-date check, which is currently applied only for a third build, because `.docker` folder is created in build directory after the first one, thus changing input for the `dockerBuild`.

Example of log from the second build:
```
Task ':analytics:computation-service:dockerBuild' is not up-to-date because:
  Input property 'inputDir' file /home/ivan/IdeaProjects/computation-service/build/.docker has been added.
  Input property 'inputDir' file /home/ivan/IdeaProjects/computation-service/build/.docker/analytics_computation-service_dockerBuild-imageId.txt has been added.
```

Also, there is a refactoring for `buildLayers` task which clearly defines its' config and makes its' cache key consistent when the `--build-cache` enabled. Thus, with some (tricky) configuration of  `dockerBuild`, it is possible to employ gradle build cache mechanism to skip the `dockerBuild` entirely, which is a great improvement for multi-modules projects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,6 @@
 // First, apply the publishing plugin
 plugins {
+  id "idea"
   id "com.gradle.plugin-publish" version "0.12.0"
   id "java-gradle-plugin"
   id "maven-publish"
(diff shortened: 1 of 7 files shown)


graemerocher:
@ivyazmitinov could you do the commits with the same GitHub account as the PR? Otherwise the CLA check won't pass

ivyazmitinov:
@graemerocher 
Done

ilopmar (approved):

Issue #198, opened 2021-05-06
`run --continuous` parameter `micronaut.io.watch.paths` uses hardcoded `"src/main"` srcDir


garkin:
`"src/main"` is hardcoded, and not deducted.
https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/c2d3b74389783d792cbc3d5885ccffa3b0fe5936/src/main/java/io/micronaut/gradle/MicronautApplicationPlugin.java#L127
But 11 lines below it's deducted using API. 
https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/c2d3b74389783d792cbc3d5885ccffa3b0fe5936/src/main/java/io/micronaut/gradle/MicronautApplicationPlugin.java#L138

graemerocher:
Good analysis, would you like to send a PR?

garkin:
Not sure about should i explicitly ping PR here or not, but just in case its #199

ilopmar:
Closed by #199

PR #199, opened 2021-05-07
Fix hardcoded "/src/main" srcDir for continuous restart watcher (#198)

garkin:
Fix for #198
From my tests `sourceSet.getAllSource()` contains Java, Kotlin and Resource folders. Basically, the same behavior as a former "src/main".

Stream API was used in the source before. 
`import java.util.stream.Collectors` is new here, though already used by the `micronaut-core`.

`sourceSets.findByName("main");` is `@Nullable` from the source code. Not sure about the correct behavior here, just added a conditional guard, as in the nearest `configureLogging()` method. Lack of `"main"` sound like very exotic and will break some not related parts of the code anyway.

I tried to add test around it, but it looks like there are no simple way to interrupt `--continuous` task.

### DIFF
--- a/src/main/java/io/micronaut/gradle/MicronautApplicationPlugin.java
+++ b/src/main/java/io/micronaut/gradle/MicronautApplicationPlugin.java
@@ -17,6 +17,7 @@ import org.gradle.api.tasks.*;
 
 import java.io.File;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


ilopmar (approved):

ilopmar:
Thanks!

PR #215, opened 2021-06-07
Create Micronaut GraalVM plugin and generate GraalVM resources file

ilopmar:
This PR creates a new MicronautGraalVM plugin (that is automatically applied by Micronaut Application plugin) that generates GraalVM `resource-config.json` automatically.

After this and the same [PR for the Maven plugin](https://github.com/micronaut-projects/micronaut-maven-plugin/pull/184) are merged, we can remove the code that does this from GraalvmTypeElementVisitor in core (for Micronaut 3.0.x)

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,21 @@ dependencies {
 }
 ```
 
+
+## Micronaut GraalVM Plugin
+
+```groovy
+plugins {
+  id "io.micronaut.graalvm" version "{version}"
+}
+```
+
+The [Micronaut graalvm plugin](https://plugins.gradle.org/plugin/io.micronaut.graalvm) is applied automatically by the
+[Micronaut application plugin](https://github.com/micronaut-projects/micronaut-gradle-plugin#micronaut-application-plugin) (see below)
+and it provides tasks to generate a GraalVM native image and also creates the GraalVM `resource-config.json` automatically
+with all the resources from the application.
+
+
 ## Micronaut Application Plugin
 
 ```groovy
@@ -438,7 +453,7 @@ You can add Gradle Shadow plugin so when running `./gradlew assemble` a runnable
 ```groovy
 plugins {
     ...
-    id "com.github.johnrengelman.shadow" version "6.0.0"
+    id "com.github.johnrengelman.shadow" version "7.0.0"
     ...
 }
 ```
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ plugins {
 }
 ```
 
-The [Micronaut graalvm plugin](https://plugins.gradle.org/plugin/io.micronaut.graalvm) is applied automatically by the
+The [Micronaut GraalVM plugin](https://plugins.gradle.org/plugin/io.micronaut.graalvm) is applied automatically by the
 [Micronaut application plugin](https://github.com/micronaut-projects/micronaut-gradle-plugin#micronaut-application-plugin) (see below)
 and it provides tasks to generate a GraalVM native image and also creates the GraalVM `resource-config.json` automatically
 with all the resources from the application.
--- a/src/main/java/io/micronaut/gradle/graalvm/GenerateResourceConfigFile.java
+++ b/src/main/java/io/micronaut/gradle/graalvm/GenerateResourceConfigFile.java
@@ -15,9 +15,7 @@
  */
 package io.micronaut.gradle.graalvm;
 
-import com.bmuschko.gradle.docker.shaded.com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
-import com.bmuschko.gradle.docker.shaded.com.fasterxml.jackson.databind.ObjectMapper;
-import com.bmuschko.gradle.docker.shaded.com.fasterxml.jackson.databind.ObjectWriter;
+import io.micronaut.gradle.graalvm.internal.ResourceConfigJsonWriter;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.GradleException;
 import org.gradle.api.file.ConfigurableFileCollection;
@@ -30,6 +28,7 @@ import org.gradle.api.tasks.PathSensitivity;
 import org.gradle.api.tasks.TaskAction;
 
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -37,7 +36,6 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -48,7 +46,6 @@ import java.util.stream.Collectors;
 public abstract class GenerateResourceConfigFile extends DefaultTask {
 
     private static final String META_INF = "META-INF";
-    private static final String RESOURCES = "resources";
     private static final String PATTERN = "pattern";
     private static final String RESOURCE_CONFIG_JSON = "resource-config.json";
     private static final List<String> EXCLUDED_META_INF_DIRECTORIES = Arrays.asList("native-image", "services");
@@ -105,8 +102,7 @@ public abstract class GenerateResourceConfigFile extends DefaultTask {
             Files.createDirectories(outputDirectory);
             File resourceConfigFile = outputDirectory.resolve(RESOURCE_CONFIG_JSON).toFile();
             System.out.println("Generating " + resourceConfigFile.getAbsolutePath());
-            ObjectWriter writer = new ObjectMapper().writer(new DefaultPrettyPrinter());
-            writer.writeValue(resourceConfigFile, Collections.singletonMap(RESOURCES, resourceList));
+            ResourceConfigJsonWriter.generateJsonFile(resourceList, new FileOutputStream(resourceConfigFile));
 
         } catch (IOException e) {
             throw new GradleException("There was an error generating GraalVM resource-config.json file", e);
@@ -155,14 +151,10 @@ public abstract class GenerateResourceConfigFile extends DefaultTask {
     }
 
     private static Map<String, String> mapToGraalResource(String resourceName) {
-        Map<String, String> result = new HashMap<>();
-
         if (resourceName.contains("*")) {
-            result.put(PATTERN, resourceName);
+            return Collections.singletonMap(PATTERN, resourceName);
         } else {
-            result.put(PATTERN, "\\Q" + resourceName + "\\E");
+            return Collections.singletonMap(PATTERN, "\\Q" + resourceName + "\\E");
         }
-
-        return result;
     }
 }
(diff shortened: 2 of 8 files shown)


src/main/java/io/micronaut/gradle/graalvm/MicronautGraalPlugin.java
```diff
@@ -135,6 +161,58 @@ public void apply(Project project) {
                     nativeImageTestTask.setDescription("Runs tests against a native image build of the server. Requires the server to allow the port to configurable with 'micronaut.server.port'.");
             })));
 
+            tasks.withType(AbstractCompile.class, compileTask -> {
+                compileTask.doLast(task -> {
```
ilopmar, line 165: I needed to hook into this task instead of `ProcessResources` because with Kotlin + OpenAPI the files are generated after the `processResources` task. With Java it works as expected using `ProcessResources` task.

The same also happened in the Maven plugin and I hooked into the `prepare-package` phase.

melix: This should probably not be implemented as a task action, and probably not on `AbstractCompile` either. Basically this would imply that compilation would generate resources, which means too much work for consumers which do not need those resources.

Ideally this should be a task instead. Would you mind explaining what this is doing an what you're trying to do?

ilopmar: I was looking forward for you to review this :-)

We need to generate GraalVM `resource-config.json` file for declaring all the resources that need to be included in the native image. 
In this same PR I "converted" `MicronautGraalPlugin` to a real plugin so users could apply it manually in case they don't use our Micronaut Application plugin.

The best scenario is that this code is only executed when creating the application fatjar or when running the `nativeImage`, `buildDockerNative`, and related tasks; i.e: only when is actually needed, as you said.

graemerocher: @melix let's discuss this tomorrow in a call

melix:
I pushed some changes, but I'd be more comfortable if it was tested against a Micronaut version which removes the AP generator. I'll try to see if I can do this easily.

ilopmar:
Thanks @melix, I'll try it locally with such version and the different test applications I was using.

src/main/java/io/micronaut/gradle/graalvm/MicronautGraalPlugin.java
```diff
@@ -135,6 +137,14 @@ public void apply(Project project) {
                     nativeImageTestTask.setDescription("Runs tests against a native image build of the server. Requires the server to allow the port to configurable with 'micronaut.server.port'.");
             })));
 
+            TaskProvider<GenerateResourceConfigFile> generateResourceConfig = configureResourcesFileGeneration(project, tasks);
+            tasks.withType(NativeImageTask.class).configureEach(nativeImage -> {
+                // This isn't great. Ideally the configuration file directories should be an input
+                // of the native image task directly, not something we patch
+                DirectoryProperty outputDirectory = generateResourceConfig.get().getOutputDirectory();
+                nativeImage.getInputs().files(generateResourceConfig);
+                nativeImage.args("-H:ConfigurationFileDirectories="+outputDirectory.get().getAsFile().getAbsolutePath());
```
ilopmar, line 146: This way of setting the arguments in the native image task doesn't work and the directory is not included in the native image. I've added `--verbose` to my test application and I can't see the directory included.

I've tried as a workaround defining manually the arg in my app and it works. Also I can see the directory being included when generated the native image.

```
nativeImage {
    imageName("basic-app")
    args = [
        "--verbose",
        "-H:ConfigurationFileDirectories=/home/ivan/workspaces/micronaut-graal-tests/micronaut-basic-app/build/generated/resources/graalvm"
    ]
}
```

graemerocher: @melix can we include it in the `META-INF/native-image/resource-config.json` inside the JAR directly? This way the JAR has the config and arguments are not needed

melix: I see, that's why I don't like how the current `NativeImage` task is setup. It basically erases whatever is configured, which isn't good, it should be more explicit. Do you have any guidance what I could use to test the actual generated image?

graemerocher: When you run native image with `--verbose` it spits out the computed arguments from all of the inputs (config files, explicit arguments etc.) I guess that is the only way today that I know of.

melix: @graemerocher that should probably work, yes!

melix: @ilopmar I have pushed another change which should fix this and added a test, which passed locally: https://scans.gradle.com/s/zojwohjssznwu/tests#:-test-io.micronaut.gradle-micronautgraalpluginspec-generate-graalvm-resource-config.json-with-openapi-and-resources-included-0

I didn't use the `jar` approach because we should avoid it if possible, but if you see other problems I'm happy to reconsider.

README.md
```diff
@@ -119,6 +119,21 @@ dependencies {
...
+
+## Micronaut GraalVM Plugin
+
+```groovy
+plugins {
+  id "io.micronaut.graalvm" version "{version}"
+}
+```
+
+The [Micronaut graalvm plugin](https://plugins.gradle.org/plugin/io.micronaut.graalvm) is applied automatically by the
```
graemerocher, line 131: probably should use the correct casing here ie `GraalVM`

src/main/java/io/micronaut/gradle/graalvm/GenerateResourceConfigFile.java
```diff
@@ -0,0 +1,168 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.gradle.graalvm;
+
+import com.bmuschko.gradle.docker.shaded.com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
+import com.bmuschko.gradle.docker.shaded.com.fasterxml.jackson.databind.ObjectMapper;
+import com.bmuschko.gradle.docker.shaded.com.fasterxml.jackson.databind.ObjectWriter;
```
graemerocher, line 20: Do we want to use shaded dependencies here from the Docker plugin?

ilopmar:
The new change works and the `resource-config.json` is now included in the native image.

I've also tried the gradle-git plugin as we've discussed this morning (from https://guides.micronaut.io/latest/adding-commit-info.html) and the generated file is not included in the `resource-config.json`.

I can see that the plugin task is executed before ours, tho:

image

The `git.properties` file is generated in `build/resources/main` so maybe we need to define that as an input directory (but not sure if this is the right way to fix it)

melix:
> The `git.properties` file is generated in build/resources/main so maybe we need to define that as an input directory (but not sure if this is the right way to fix it)

Can you point me at what is generating this file, or where it's configured where this file is generated? It should indeed be declared as an input.

src/main/java/io/micronaut/gradle/graalvm/GenerateResourceConfigFile.java
melix: Probably not. However I'm not super keen on adding a JSON dependency just for this, especially if we get this migrated to an official GraalVM plugin later. I would be tempted to keep this as is for the first release, then we can revisit if we don't manage to get this shipped on time, wdyt?

ilopmar:
> > The `git.properties` file is generated in build/resources/main so maybe we need to define that as an input directory (but not sure if this is the right way to fix it)
> 
> Can you point me at what is generating this file, or where it's configured where this file is generated? It should indeed be declared as an input.

The Gradle-Git plugin is the one that generates the file. You can reproduce it easily downloading the this guide https://guides.micronaut.io/latest/adding-commit-info-gradle-java.html.

melix:
> The Gradle-Git plugin is the one that generates the file. You can reproduce it easily downloading the this guide https://guides.micronaut.io/latest/adding-commit-info-gradle-java.html.

Alright so there are 2 ways we can support this:

1. add some hook into the Micronaut plugin, to detect the application of the Git properties plugin and do the wiring automatically
2. change the userguide documentation to explain how to include the properties file, as today, but declared as an input

I would lean towards 2 since introducing coupling to a 3rd party plugin like this doesn't sound quite correct. Basically, we'd need to document to do this instead:

```groovy
tasks.named("generateResourceConfigFile") {
   resourceDirectories.from(tasks.findByName('generateGitProperties').output.map { it.getAsFile().parent })
}
```

melix:
Note that it would probably be nice to have "companion plugins" which auto-configure Micronaut in a useful way when we detect those (for example to setup the above), but it's not for tomorrow :)

src/main/java/io/micronaut/gradle/graalvm/GenerateResourceConfigFile.java
bmuschko: I wouldn't rely on those shaded imports as Docker Java now split up their library into smaller chunks. As a result, the footprint of transitive dependencies will be smaller. In one of the latest versions of Docker Java, we started to depend on less of those dependencies from the Docker Gradle plugin. I am not sure about their long-term plans but Jackson might go away.

graemerocher:
I think documenting how to extend is fine from my perspective

ilopmar:
I was trying the same with the Maven plugin and looking for additional resources in the "output" directory instead of the `META-INF` directory inside it, makes it work.

I think that, as the Git plugin put the generated resources in `build/resources/main` we should include that directory. It's not that it generates the file in some random directory, but in the default one. This way both plugins will have the same features and users won't need to do anything manually.

melix:
>  think that, as the Git plugin put the generated resources in build/resources/main we should include that directory

This isn't a standard Gradle output directory, tbh. There are different Git plugins and we can't really support them all, can we? In practice, this plugin also systematically adds `git.properties` in the jar file which also breaks caching. I'm not sure we should recommend this.

src/main/java/io/micronaut/gradle/graalvm/GenerateResourceConfigFile.java
melix: Right. So given the current contents of the file I think we should just go with hand written JSON, that should be pretty straightforward.

ilopmar:
> This isn't a standard Gradle output directory, tbh. There are different Git plugins and we can't really support them all, can we? In practice, this plugin also systematically adds `git.properties` in the jar file which also breaks caching. I'm not sure we should recommend this.

Ok, fair enough. Then we'll document how to include this in the Micronaut Guide for Gradle/Maven.

graemerocher (approved):

Issue #161, opened 2021-03-15
Consider to add a specific USER to the Dockerfile


saw303:
The Micronaut Gradle plugin sets the `WORKDIR` to `/home/app` by default but runs the application in the context of the `root` user.

According to the [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) of Docker manual it would be great if the plugin would start the application in the user context of user `app` instead of `root`.

```Dockerfile
//.... base image
WORKDIR /home/app
COPY build/layers/libs /home/app/libs
COPY build/layers/resources /home/app/resources
COPY build/layers/application.jar /home/app/application.jar
RUN groupadd -r app && useradd --no-log-init -r -g app app && chown -R app:app /home/app
USER app
ENTRYPOINT ["java", "-jar", "/home/app/application.jar"]
```

martinhynar:
Using Micronaut 2.5.6 and building docker image now fails with 

```
Step 6/8 : RUN addgroup app && adduser -G app app -D && chown -R app:app /home/app
 ---> Running in e4e5d7ef0028
/bin/sh: addgroup: command not found
```

I am using OracleLinux based base image and the command is `groupadd` (as suggested in the initial comment) not `addgroup`.

ilopmar:
I used `addgroup` because we don't use Oracle Linux by default and `groupadd` doesn't work with openjdk-alpine

martinhynar:
Same problem with Fedora and I assume that other distros will suffer with that too. IMO, it needs to be either configurable or somehow flexible (execute groupadd if the addgroup fails)

ilopmar:
PRs welcome :)

ilopmar:
Fixed by #221

PR #221, opened 2021-06-22
2 variants of adding dedicated group and user

martinhynar:
Referenced issue: #161

This PR adds support for base images that have either groupadd or addgroup. It check which one is available and based on that triggers command. Also, when specific group manipulation command is available, useradd has different syntax, so this is also reflected.

Tested with openjdk:16-alpine and oraclelinux:7


ilopmar:
Thanks for the contribution. I'll release a new patch version with this fix and I'll ping you

ilopmar:
@martinhynar fix included in `1.5.3`.

PR #227, opened 2021-06-30
Allow to pass additional instructions to be included in Dockerfile

ilopmar:
Those additional instructions will be included just before the
entrypoint.
This also removed the automatic user and group creation because it had
issues with different base images. Now with the additional instructions
users can create a user and group for their specific base image.

Fixes #224

I'm targeting master branch in this PR because this is a breaking change in the behaviour about creating a user and group to run the application.


melix (approved):
This LGTM as a hotfix. The additional instructions should probably be modeled as an distinct input rather than clearing/setting, but this can be done later.

PR #228, opened 2021-07-07
Rework the Gradle plugin to use the official GraalVM plugin

melix:
Please look at individual commits for details.


melix:
Should we have a version bump with this change?

ilopmar:
I think this PR introduces breaking changes, right? Then in that case we need a new major version. We should also upgrade the docs (at this moment `README.md`) and target Micronaut 3.1

melix:
Yes it introduces breaking changes. Where should those be documented? Also, in practice, what does it mean to "target Micronaut 3.1"?

ilopmar:
> Yes it introduces breaking changes.

Then we need to release 3.0.0 


> Where should those be documented? 

The plugin lacks of documentation other than the README file so at this moment adding docs mean updating README.md. We should probably upgrade this to have proper Asciidoctor documentation like we have in all the modules.
Having said that I don't know if you plan to "merge" this plugin and the Maven one in one common repository. If so, maybe that's the right time to add the docs.


> Also, in practice, what does it mean to "target Micronaut 3.1"?

In the rest of the modules, "targeting a specific Micronaut version" means that automatic PRs with the version update after a release is done will be sent to a specific branch in Micronaut core. In practice it means to have the right branch in here https://github.com/micronaut-projects/micronaut-flyway/blob/master/gradle.properties#L18.

The build in this plugin is so different from the modules that it doesn't do that. As you probably remember to upgrade the version of the plugin in Micronaut we need to send a PR to Starter (like https://github.com/micronaut-projects/micronaut-starter/pull/899/files). Maybe now that we are using the catalog we can improve that.

So, for this module it means that the PR to use this new version of the plugin needs to go to Starter 3.1.x branch.

melix:
> Having said that I don't know if you plan to "merge" this plugin and the Maven one in one common repository. If so, maybe that's the right time to add the docs.

I don't see any particular value in merging the repositories, unless we're starting to have common code between the two plugins. It's good that we can release distinct versions for both. I can add documentation to this one independently.

> So, for this module it means that the PR to use this new version of the plugin needs to go to Starter 3.1.x branch.

Right, so this is actually fine. It also means we could start releasing 3.x versions of this plugin _before_ Micronaut 3.1 is out (so that folks can try or fix some of the reported bugs).

melix:
@graemerocher @ilopmar I need to confirm that this won't make it into 3.1, but 3.2, do we agree?

ilopmar:
Yes, it's better to wait at least until 3.2 as long as Gradle Toolchain issue is resolved https://github.com/gradle/gradle/issues/18290 or we force the flags to behave as users expect.

graemerocher:
Yes that is fine

PR #229, opened 2021-07-09
Fix detection of GraalVM in IntelliJ IDEA

morki:
In IntelliJ IDEA, the `java.vendor.version` property is inherited from IDEA's JDK, so `nativeImage` task is reporting this error message:

```
A GraalVM SDK is required to build native images
```

The `java.vendor` property is working well with value: "GraalVM Community"

See related issues here:
- https://youtrack.jetbrains.com/issue/IDEA-271768
- https://youtrack.jetbrains.com/issue/IDEA-268520

### DIFF
--- a/src/main/java/io/micronaut/gradle/graalvm/GraalUtil.java
+++ b/src/main/java/io/micronaut/gradle/graalvm/GraalUtil.java
@@ -11,7 +11,7 @@ public final class GraalUtil {
      * @return Return whether the JVM in use a GraalVM JVM.
      */
     public static boolean isGraalJVM() {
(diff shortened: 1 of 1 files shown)


graemerocher:
@melix might be applicable to the official plugins?

melix:
The official plugin uses a different mechanism (toolchain support and/or `GRAALVM_HOME` support) so it should be fine.

melix (approved):

morki:
@graemerocher can we please merge this as a temporary fix before we will use official plugin internally? Or will it be in some near feature so this does not make sense? Thank you for feedback :)

melix (approved):

PR #245, opened 2021-09-07
Resolve #237 - Unable to use custom entrypoint in generated Dockerfile

### LINKED ISSUE #237: Unable to use custom entrypoint in generated Dockerfile
We use a custom base image for our docker containers. Our entrypoint in that container is a script, that sets up some variables and does some other manipulation on the container before running the Java command.

We would like to use the generated `Dockerfile` that this plugin creates, but are unable to find a mechanism to do this with the `MicronautDockerfile` task.

So far we have worked around this problem by using our own `Dockerfile`, but this is not ideal, because we are subject to path changes of the `buildLayers` task like what happened in https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/179.

```
FROM <company-repository>/base-images/java11:2021.0527.1643

COPY build/docker/layers/libs /home/app/libs
COPY build/docker/layers/resources /home/app/resources
COPY build/docker/layers/application.jar /home/app/application.jar

ENTRYPOINT ["/entrypoint.sh"]
```

Ideally we could define this in our `build.gradle`
```
dockerfile {
  baseImage = "<company-repository>/base-images/java11:2021.0527.1643"
  entryPoint("/entrypoint.sh")
}
```
but when we do that we get a generated `Dockerfile` with 2 entrypoint scripts

```
FROM <company-repository>/base-images/java11:2021.0527.1643
WORKDIR /home/app
EXPOSE 8080
ENTRYPOINT ["/entrypoint.sh"]
ENTRYPOINT ["java", "-jar", "/home/app/application.jar"]
```

I am happy to submit a PR with a "fix" and currently have 2 thoughts for how to achieve this:
1. Create a new variable in `MicronautDockerfile.java` something like `customEntrypoint` and if it exists, use that, and if it doesn't then use what exists today.
2. Iterate the instruction set from the base `Dockerfile` class and if an entrypoint exists skip adding the `MicronautDockerfile` entrypoint.

Personally I think option 1 is more explicit and generally lines up with the established pattern in `MicronautDockerfile.java`.

Rough example of Solution 1 implementation and result
```
dockerfile {
  baseImage = "<company-repository>/base-images/java11:2021.0527.1643"
  customEntrypoint = "/entrypoint.sh"
}
```
generates 
```
FROM <company-repository>/base-images/java11:2021.0527.1643
WORKDIR /home/app
EXPOSE 8080
ENTRYPOINT ["/entrypoint.sh"]
```

frossbeamish:
This solves #237  - Unable to use custom entrypoint in generated Dockerfile.

By specifying a `customEntrypoint` parameter to dockerfile or nativedockerfile closures the generated entrypoint will be overwritten completely by what is passed in versus concatenating multiple lists.

### DIFF
--- a/src/main/java/io/micronaut/gradle/docker/MicronautDockerfile.java
+++ b/src/main/java/io/micronaut/gradle/docker/MicronautDockerfile.java
@@ -30,8 +30,7 @@ public class MicronautDockerfile extends Dockerfile implements DockerBuildOption
     private final Property<DockerBuildStrategy> buildStrategy;
     @Input
     private final Property<String> defaultCommand;
-    @Input
-    private final ListProperty<String> customEntrypoint;
+
 
     public MicronautDockerfile() {
         Project project = getProject();
@@ -45,8 +44,6 @@ public class MicronautDockerfile extends Dockerfile implements DockerBuildOption
         this.args = objects.listProperty(String.class);
         this.exposedPorts = objects.listProperty(Integer.class)
                     .convention(Collections.singletonList(8080));
-        this.customEntrypoint = objects.listProperty(String.class)
-                    .convention(Collections.emptyList());
 
         //noinspection Convert2Lambda
         doLast(new Action<Task>() {
@@ -93,9 +90,7 @@ public class MicronautDockerfile extends Dockerfile implements DockerBuildOption
                 setupResources(this);
                 exposePort(exposedPorts);
                 getInstructions().addAll(additionalInstructions);
-                if (getCustomEntrypoint().get().size() > 0) {
-                    entryPoint(getCustomEntrypoint());
-                } else {
+                if (getInstructions().get().stream().noneMatch(instruction -> instruction.getKeyword().equals(EntryPointInstruction.KEYWORD))) {
                     entryPoint(getArgs().map(strings -> {
                         List<String> newList = new ArrayList<>(strings.size() + 3);
                         newList.add("java");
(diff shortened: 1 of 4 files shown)


frossbeamish:
@ilopmar Can you take a look at this PR when you have time. This should fix #237

ilopmar:
@frossbeamish I'll leave this to @melix

src/main/java/io/micronaut/gradle/docker/MicronautDockerfile.java
```diff
@@ -30,8 +30,7 @@
     private final Property<DockerBuildStrategy> buildStrategy;
     @Input
     private final Property<String> defaultCommand;
-    @Input
-    private final ListProperty<String> customEntrypoint;
+
```
melix: Sorry that's not what I meant. I meant that the _property name_ was wrong: it should be `entryPoint`, and its default value should be configured with the `getArgs().map...` thing.

frossbeamish: @melix with the current approach as it exists we are able to leverage the existing `entryPoint` method that exists in the base `Dockerfile` task. I worried about overloading the method with a property of the same name that does essentially the same thing being confusing and not clear to a user.

In fact my first approach to solving #237 was to use the base dockerfile `entryPoint` method, it just generated the resulting Dockerfile incorrectly because that task doesn't have any logic to limit you to one entrypoint.

melix: No worries, let me take a look at this.

melix:
Closed in favor of #256 (rebased/cleaned up)

PR #264, opened 2021-10-04
[gradle-plugin] Update common files for branch master

micronaut-build:
Update common files


melix:
Do not merge this PR as is or it would break deployment to the plugin portal.

PR #278, opened 2021-10-26
Introduce proper documentation

melix:
This commit adds proper documentation for the Gradle plugin: instead
of using a README the docs now use Asciidoc and should be published
to GitHub pages.


sdelamo (commented):
I have created several styling PRs against this PR's base. 

[Links to Gradle Plugin Portal](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/281)
[Margin bottom copy to clipboard](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/282)
[links to new youtube channel](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/283)
[fix: broken foundation image](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/284)
[remove list style type from table of contents](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/285)
[makes the table of contents area narrower](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/286)
[break line to improve readability](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/287)
[better margins](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/288)
[increase max-width area](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/289)
[css to style asciidoc tips](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/290)

Moreover, I have created a [PR](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/291) which uses remote CSS and Javascript files outside of this repository to avoid duplication of those files in every repository. 


I assume that, apart of moving documentation outside of the README file, the goal is to have a first stone about how we could do documentation for every module in the future. 

Some questions

- will `Themer` be shipped as part of Micronaut docs instead of being in `buildSrc` of each module. 
- Could logic `io.micrnaut.internal.build.documented.gradle` be moved to a central place insated of having it in each module. 
- Could `template.html` be moved to a central place?


We need to support configuration reference and javadoc. Any plans for this? This module will not have configuration reference but I assume that we should publish javadoc of it.

We need to support a numbered table of contents.

melix:
Thanks @sdelamo !

> I assume that, apart of moving documentation outside of the README file, the goal is to have a first stone about how we could do documentation for every module in the future.

I'm not quite sure yet. Basically this project is a bit special, similar to the guides repo, in the sense that it doesn't define a Micronaut module but something different (a Gradle plugin here) so some of the assumptions of the current docs infra do not stand.

> will Themer be shipped as part of Micronaut docs instead of being in buildSrc of each module.

I think if we find this setup useful, it would be indeed good to have this code extracted. Should it be in `micronaut-build` or not, is a question that we have to answer. I understand that the docs generation used to be in `micronaut-docs`, then it was moved in `micronaut-build`. This causes some issues, so maybe we should move everything back to a `micronaut-docs` project, with a distinct lifecycle from `micronaut-build`. I have no strong opinion yet because I'm still trying to wrap my head around the setup.

> Could template.html be moved to a central place?

Same answer, but if we move the `Themer`, we should move everything needed to generate docs into the same repo.

> We need to support configuration reference and javadoc. Any plans for this? This module will not have configuration reference but I assume that we should publish javadoc of it.

Configuration reference and javadocs are supported if we use the current docs plugin, which is applied automatically with the `micronaut-build` plugin. The goal of this PR is _not_ to replace those, even if it can help and be inspiration for a new way of generating docs. Regarding javadocs, I am not sure we _need_ to publish them (well, they are already published as a jar but I assume you're talking about publishing the rendered result alongside the userguide). If we do it needs to look similar to what we have on other modules.

Question: does #291 include all the changes I merged from your other PRs?

ilopmar:
@sdelamo can we apply the same styles and look and feel than the rest of the micronaut docs? Even with all your new changes this documentation looks like a Micronaut Guide. Unless, of course, we plan to change the look and feel in all the docs to look like the guides...

melix:
As discussed during the coordination meeting I'm merging this PR and will follow up with javadocs.

PR #280, opened 2021-10-28
Integrate with Micronaut AOT

melix:
This commit introduces a new `io.micronaut.aot` plugin which integrates
with the Micronaut AOT module. If applied, this plugin will configure
the build to:

1. register some `prepareXXXOptimizations` tasks for optimizing apps
for JIT and native
2. register new `optimizedRun` task which executes a Micronaut app
with the result of the AOT optimizations
3. register new `nativeOptimizedRun` task which executes a Micronaut
app natively (with GraalVM) with the result of the native optimizations

The AOT optimization configuration is exposed with an `aot` sub-extension
in the `micronaut` extension. Currently there's no opt-in because all
the tasks which are being created are isolated from the standard tasks:
it just makes the AOT optimizations visible, without interfering with
the standard execution.

Future work may include an option to replace the standard tasks with
the optimized versions, but it will require more work to get there (in
particular, we need to think what it means for test execution).


melix:
Well the issue is similar to testing in native mode: in order to gather the list of tests, test must be executed first in "JVM" mode so that we can introspect what's going to run and build a list for the "native" execution. Here it's similar: just analyzing the classpath isn't enough to understand what `MicronautTest` is going to be called, nor how the application contexts are going to be executed. So we'd have to run the tests in a first step, possibly with a custom JUnit engine which only triggers MicronautTest configuration, then use this information somehow in AOT.

Alternatively, we need to figure out a way to change `micronaut-test` so that it uses `ContextConfigurer`. However, this is a semantic change since `ContextConfigurer` is unconditional, and that `@MicronautTest` isn't.

melix:
@graemerocher I have created a `3.0.x` maintenance branch, and `master` is now `3.1.x`. Are you ok if I merge this and do follow-ups as individual PRs?

graemerocher:
fine by me

PR #301, opened 2021-11-08
Fix links to Gradle docs

ilopmar:
Before this PR, links to Gradle docs are wrong because they include duplicated `7.2/userguide`, like `https://docs.gradle.org/7.2/userguide/7.2/userguide/toolchains.html`.


melix:
Oh, good catch. Not the first time I wished the asciidoc Gradle plugin has an embedded link checker!

melix (approved):

PR #304, opened 2021-11-10
remove links from navigation but api reference

sdelamo:
I don't think the changes done here are consistent with our navigation in other modules' documentation 

https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/295

Screenshot 2021-11-10 at 07 16 17

Example of other modules: 

Screenshot 2021-11-10 at 07 16 44

### DIFF
--- a/src/docs/asciidoc/template/template.html
+++ b/src/docs/asciidoc/template/template.html
@@ -37,72 +37,10 @@
 
                 <ul class="navbar-nav ml-auto">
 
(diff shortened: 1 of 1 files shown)


melix:
I don't have a strong opinion on this. We can remove the links, or we can switch to the template used by the other projects for documentation.

sdelamo:
At least remove the links. What we currently have does not work at all.

melix (approved):

PR #320, opened 2021-11-26
First batch of documentation of AOT integration

melix:


src/docs/asciidoc/index.adoc
```diff
@@ -815,6 +816,257 @@ and it provides tasks to generate a GraalVM native image and also creates the Gr
...
+serviceloading.rejected.impls = com.Misc,org.Bar
+
+# Converts YAML configuration files to Java configuration
+yaml.to.java.config.enabled = true
+
+# Precomputes property sources at build time
+sealed.property.source.enabled = true
+----
+
+For native images, it is important to always have the `graalvm.config.enabled` option set to `true`, otherwise the AOT optimizations will not be loaded.
```
ilopmar, line 998: How can we set that using the DSL?

src/docs/asciidoc/index.adoc
```diff
@@ -815,6 +816,257 @@ and it provides tasks to generate a GraalVM native image and also creates the Gr
...
+    }
+}
+----
+
+NOTE: You can provide both a configuration file and `aot` DSL optimizations.
+The configuration will be merged, by reading the file first, then using the DSL options.
+
+If you want to know about all possible optimizations, you can run the `createAotSampleConfigurationFiles` which will generate a couple of sample files:
+
+The `build/generated/aot/samples/jit/jit.properties` will contain the optimizations which are relevant to an application running in the regular Java virtual machine, for example:
```
ilopmar, line 925: It good be nice to show the file created when running the task. Here in the docs it's explained but it is more helpful to have it when running the task. Something lilke:

```
$ gw createAotSampleConfigurationFiles                                  
...

> Task :createAotJitSample
Created build/generated/aot/samples/jit/jit.properties <----- 

> Task :createAotNativeSample
___ And the same here__

> Task :createAotSampleConfigurationFiles

___ And here__
```

ilopmar (changes requested):

src/docs/asciidoc/index.adoc
melix: You don't, the plugin does it automatically.

ilopmar: :+1:

src/docs/asciidoc/index.adoc
melix: In general I'm not a big fan of verbose builds. Having something which dumps everything it does. For example `compileJava` doesn't output "wrote class files in ...". Now for this particular case, since it's tasks whose sole goal is to generate a sample, why not. But certainly not duplicated in both the tasks which create the samples and the one which is the lifecycle task :)

ilopmar: Yes, I agree that for common tasks it doesn't make sense but for some of them it helps users a lot

ilopmar (approved):

PR #349, opened 2021-12-21
Add support for optimizing lambdas

melix:
This commit adds initial support for optimizing Micronaut libraries,
in particular functions. For now tests limited to lambdas, but
deployment is **not** tested yet.

### DIFF
--- a/samples/aot/aws-function/build.gradle
+++ b/samples/aot/aws-function/build.gradle
@@ -1,0 +1,53 @@
+plugins {
+    id("io.micronaut.library") version "3.1.1-SNAPSHOT"
+    id("io.micronaut.aot") version "3.1.1-SNAPSHOT"
+    id("com.github.johnrengelman.shadow") version "7.1.0"
+}
+
+version = "0.1"
+group = "example.micronaut"
+
+repositories {
+    mavenCentral()
+}
+
+micronaut {
+    version = "3.2.0"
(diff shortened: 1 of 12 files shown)


graemerocher:
We need variations of `buildLambdaZip` and `buildNativeLambda` that build the optimized versions. See https://www.youtube.com/watch?v=76JPU9cyuLk for demos of usage

melix:
Will take a look, thanks. One question I have is regarding deployment of the native version. Currently the function is a _library_, which means that `native-image` will generate a `.so`. This means, as far as I understand, that we need a custom runtime to be able to run this, no? All guides I have found so far workaround the problem by creating an _application_ instead.

graemerocher:
A custom runtime is used to deploy Lambda with native image. That is why we set this here https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/1706100deb116b89298bb31f3c2093afa7bdf669/src/main/java/io/micronaut/gradle/graalvm/MicronautGraalPlugin.java#L75

melix:
But isn't this custom runtime needing an actual `main()` method, so an `application`?

graemerocher:
Correct

melix:
With AOT 1.0.0-M6 this should now work properly.

graemerocher (approved):

PR #355, opened 2021-12-22
Make the created jar tasks a dependency of assemble

melix:
So that this lifecyle task also builds optimized jars


melix:
@graemerocher @ilopmar do you have an idea why GH gets stuck on waiting for `build(17)`?

ilopmar:
> @graemerocher @ilopmar do you have an idea why GH gets stuck on waiting for `build(17)`?

It's because the project is configured to check the `build (17)` task, but this still uses JDK16. You need to update the GitHub workflow to use JDK17.

P.S: If everything is ok I can merge the PR for you.

melix:
I see, thanks, yes, please merge the PR, I'll fix the workflow in another one.

PR #360, opened 2022-01-05
Refactor the plugin architecture

melix:
This pull request refactors the Gradle plugin in order to make it possible to build libraries without adding GraalVM support, or building applications without docker or GraalVM support (nor applying the Eclipse plugin).

There are a couple of new user facing plugins:

- `io.micronaut.minimal.library` : lets the user build a Micronaut Library, but doesn't have support for GraalVM
- `io.micronaut.minimal.application`: lets the user build a Micronaut application, but doesn't apply the GraalVM or Docker plugins.

For example, instead of using the `io.micronaut.application` if you don't need the Docker support but only the GraalVM support, you can use:

```
plugins {
    id 'io.micronaut.minimal.application'
    id 'io.micronaut.graalvm'
}
```


src/main/java/io/micronaut/gradle/MicronautComponentPlugin.java
```diff
@@ -0,0 +1,334 @@
+/*
+ * Copyright 2003-2021 the original author or authors.
```
burtbeckwith, line 2: 2022?

melix: Oh, time to reprocess all headers :)

ilopmar (approved):
LGTM :+1:

PR #366, opened 2022-01-14
ARM Lambda native images can be built on M1 machines

timyates:
Prior to this, when `buildNativeLambda` is executed on an M1 Apple machine, it pulls the
ARM version of amazonlinux:latest, and then installs the x86 version of GraalVM.

This obviously fails.

This change detects if the amazonlinux docker image is running on an ARM architecture,
and the fetches the correct GraalVM JDK from GitHub.

It does not allow cross-compilation of x86 Lambda images on M1 hardware.  Currently the
docker plugin we use does not allow buildx, nor setting the platform when executing docker

I'm not 100% it will be possible even with a later image.

I have tested the resulting ARM lambda on AWS when build on an M1 mac.

I am going to test the native image from an x86 Mac as well. I'm relying on GitHub to test
Linux and Windows


timyates:
Currently trying to track down a JNA issue (we're pulling 5.6.0 into the build, which doesn't work on ARM)

PR #367, opened 2022-01-14
Fix Lambda native image building on Apple M1

timyates:
Prior to this, when `buildNativeLambda` is executed on an M1 Apple machine, it pulls the
ARM version of amazonlinux:latest, and then installs the x86 version of GraalVM.

This obviously fails.

This change detects if the amazonlinux docker image is running on an ARM architecture,
and the fetches the correct GraalVM JDK from GitHub.

**Currently this does not perform cross-compilation of x86 Lambda images on M1 hardware.  Currently the
docker plugin we use does not allow buildx, nor setting the platform when executing docker**

I'm not 100% it will be possible even with a later image.

I have tested the resulting ARM lambda on AWS when build on an M1 mac, and an x86 lambda when built on an intel mac.

Something needs to be done with a JNA 5.6.0 library that's getting transitively pulled in to the project when the plugin is applied

### DIFF
--- a/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -356,7 +356,9 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
             runCommand("yum install -y gcc gcc-c++ libc6-dev zlib1g-dev curl bash zlib zlib-devel zlib-static zip tar gzip");
             String jdkVersion = getJdkVersion().get();
             String graalVersion = getGraalVersion().get();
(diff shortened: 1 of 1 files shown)


timyates:
After talking with @melix, we need to think about building x86 images on arm hardware

I tried, but hit segfaults, which I have raised with the native-image slack channel

timyates:
We can set

```
dockerfileNative {
    graalArch.set("aarch64")
}
```

With

```
    runtime("lambda_provided")
```

And add this to the top of the build

```
buildscript {
    dependencies {
        classpath("com.github.docker-java:docker-java-transport-httpclient5:3.2.12") {
            because("M1 macs need a later version of JNA")
        }
    }
}
```

To build an arm lambda inside a docker container

PR #374, opened 2022-01-26
Cc/publish multiple plugins

melix:


ljacomet:
The publication with the same plugin ID but using different GAV coordinates for the plugin JAR will effectively trigger a manual validation step on the plugin portal.
You could reach out right before making that first release so it gets approved in a timely fashion, although we now are much better on that front 😉

melix:
Right, that's what I feared. Since we release from CI, this may be tricky if there's an approval error. Do you see anything obvious that I could be missing, to reduce the risks?

PR #378, opened 2022-01-28
Fix native docker file using wrong paths under Windows

### LINKED ISSUE #358: Task dockerBuildNative produces wrong path in DockerfileNative when it is executed under Windows
### Expected Behavior

Executing micronaut gradle plugin task `dockerBuildNative` should create a native compiled docker image.

This is working under WSL2 Ubuntu but not in Windows. See details below:

Command: gradle :app:dockerBuildNative

image

### Actual Behaviour

It stops in Step RUN native-image because there is a path issue to the referenced \home\app\application.jar instead of /home/app/application.jar. It seems that the produced DockerfileNative has path issues.

This happens when running the same command under CMD in the Windows terminal. See details below:

Command: gradle :app:dockerBuildNative

image

### Steps To Reproduce

1. Open CMD
2. gradle/gradlew :<application>:dockerBuildNative

### Environment Information

- Windows 10 (Windows 10 Pro, 21H1, 19043.1415)
- Gradle/Gradlew 7.2
- openjdk 17 2021-09-14
  - OpenJDK Runtime Environment (build 17+35-2724)
  - OpenJDK 64-Bit Server VM (build 17+35-2724, mixed mode, sharing)

### Example Application

_No response_

### Version

3.2.0

melix:
Fixes #358

### DIFF
--- a/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -464,13 +464,8 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
         }
         List<String> commandLine = new ArrayList<>();
         commandLine.add("native-image");
-        commandLine.addAll(new NativeImageCommandLineProvider(
-                getProviders().provider(() -> options),
-                getProviders().provider(() -> false),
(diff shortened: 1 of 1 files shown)


melix:
I have verified the fix works in a Windows VM.

Issue #279, opened 2021-10-26
When Deploying to AWS Lambda as GraalVM native image, allow modification of baseImage. 


runscinch:
### Feature description

The docker file generated by `./gradlew buildNativeLambda` doesn't allow for modification of the base image. I have the requirement to append a corporate url to the image being used. For example I need
`FROM amazonlinux:latest AS graalvm`

to be `FROM internal.proxy.com/amazonlinux:latest AS graalvm`

I see that value is currently hardcoded. 

Thank you!

PR #383, opened 2022-02-03
Updating to allow override of base Image for lambda GraalVM

runscinch:


runscinch:
Related to this issue: https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/279

This change would allow the base image to be updated from default for Lambda runtime.

melix (commented):
Would you mind adding a test for this?

runscinch:
@melix sure, would you prefer I create a new class or add to an existing class? Wasn't sure the best fit.

melix:
Adding to `io.micronaut.gradle.LambdaNativeImageSpec` probably makes sense, thanks!

melix (approved):

Issue #365, opened 2022-01-13
Micronaut Serverless Function is not build with Graal Native support on Apple Silicon


tamershahin:
### Expected Behavior

I expect when using Apple Silicon MacBook that the command:

`$ ./gradlew buildNativeLambda` will output a valid image and .zip file with the function 

### Actual Behaviour

I receive the error:

``` nown> Dockerfile: ../test/micronautguide/build/docker/native-main/DockerfileNative'…
Building image…
Preparing build context archive…
[==================================================>]78/78 files
Done

Sending build context to Docker daemon…
[==================================================>] 19.68MB
Done

Step 1/27 : FROM amazonlinux:latest AS graalvm
 ---> 377da0386f50
Step 2/27 : ENV LANG=en_US.UTF-8
 ---> Using cache
 ---> b3f0a3337e46
Step 3/27 : RUN yum install -y gcc gcc-c++ libc6-dev zlib1g-dev curl bash zlib zlib-devel zlib-static zip tar gzip
 ---> Using cache
 ---> 336cb10c6e4f
Step 4/27 : RUN curl -4 -L https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.3.0/graalvm-ce-java11-linux-amd64-21.3.0.tar.gz -o /tmp/graalvm-ce-java11-linux-amd64-21.3.0.tar.gz
 ---> Using cache
 ---> 6b6fb6b0202d
Step 5/27 : RUN tar -zxf /tmp/graalvm-ce-java11-linux-amd64-21.3.0.tar.gz -C /tmp && mv /tmp/graalvm-ce-java11-21.3.0 /usr/lib/graalvm
 ---> Using cache
 ---> 9d9451ea1cda
Step 6/27 : RUN rm -rf /tmp/*
 ---> Using cache
 ---> e5c5dfae4002
Step 7/27 : RUN /usr/lib/graalvm/bin/gu install native-image
 ---> Running in 7c8bba443427
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
Error response from daemon: The command '/bin/sh -c /usr/lib/graalvm/bin/gu install native-image' returned a non-zero code: 255
Failed to deploy '<unknown> Dockerfile: ../test/micronautguide/build/docker/native-main/DockerfileNative': Can't retrieve image ID from build stream
``` 

### Steps To Reproduce

step 1:

[...]

PR #384, opened 2022-02-03
Bump version of Gradle Docker Plugin

dzou:
This bumps the version of the Gradle Docker Plugin to `7.1.0` in order to accommodate compilation for Apple M1 processors. Some users have had trouble running `dockerBuild` on Apple M1 devices.

See:
* https://github.com/docker-java/docker-java/issues/1779
* https://github.com/bmuschko/gradle-docker-plugin/issues/1035

Following the [suggestion here](https://github.com/bmuschko/gradle-docker-plugin/issues/1035#issuecomment-1023275441), upgrading to the latest version of gradle-docker-plugin seems to resolve the docker issues with Apple M1 processors.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 kotlin = "1.4.32"
-docker = "6.7.0"
+docker = "7.1.0"
 diffplug = "3.33.2"
 shadow = "7.1.0"
 groovy = "3.0.9"


dzou:
@melix -- What are your thoughts about this change? I think upgrading the version will unblock many users attempting to Docker build Micronaut projects on Apple M1 machines.

melix:
This is actually discussed in https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/365

We can't upgrade for now because we still have to support running on JDK 8. This will be doable once Micronaut 4 is out.

dzou:
Ahh okay. Is some way for us to override the version of `gradle-docker-plugin` pulled in by micronaut-plugin from our gradle build file? Sorry I am a bit inexperienced with how one might achieve this.

dzou:
I think this might work as a workaround for now.. but not completely sure.

```
    buildscript {
        dependencies {
            constraints {
                classpath ('com.bmuschko:gradle-docker-plugin:7.1.0')
            }
        }
    }
```

melix:
Alternatively you can do this, in your `build.gradle` file:

```
id( "com.bmuschko.docker-remote-api" ) version "7.2.0" apply false
```

and then target x86 for your image:

```
dockerBuild {
    platform.set( "linux/amd64" )
}
```

which is, I think, more likely what you need (that is, build an x86 image on M1)

dzou:
Thanks. I'll just close this for now then.

PR #389, opened 2022-02-16
Upgrade to GraalVM 22.0.0.2 by default

melix:


melix:
Deferring the change to a later release, because of https://github.com/graalvm/container/issues/49 (which also happens with x86)

graemerocher (approved):

melix:
@sdelamo builds are passing now. Should we target 3.3.2 for this?

PR #390, opened 2022-02-19
[gradle-plugin] Update common files for branch master

micronaut-build:
Update common files


melix:
@alvarosanchez can you add static analysis for this repo?

alvarosanchez:
Done. You still need to add the quality reporting plugin to the root project I guess.

melix:
The quality check plugin is probably not compatible with additional test tasks: https://github.com/micronaut-projects/micronaut-gradle-plugin/runs/5408038322?check_suite_focus=true#step:7:183

alvarosanchez:
It doesn't seem that is the issue, but rather not picking up [this properties](https://github.com/micronaut-projects/micronaut-build/blob/master/src/main/java/io/micronaut/build/MicronautQualityReportingAggregatorPlugin.java#L42)

Maybe it needs the build common plugin?

At this point, it's possibly not worth it, and you can simply configure manually the `org.sonarqube` plugin directly to this project.

melix:
Ah right, could it be that the sonar env vars are not setup in that repo?

alvarosanchez:
No, it is passed: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/390/files#diff-a5e11e7bedd328777f4897395a711a4492d413b70b12479230470ec608d96792R57. And the secret is an organization-level secret. It's present in all our repos.

PR #413, opened 2022-03-25
refactor: extract MicronautLambdaRuntime class as a constant

sdelamo:


docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautLambdaUtils.java
```diff
@@ -0,0 +1,36 @@
...
+     * Constructor.
+     */
+    private MicronautLambdaUtils() {
+
+    }
+
+    /**
+     * Main class for Provided runtime and Application Type: Micronaut Application.
+     */
+    public static final String MICRONAUT_LAMBDA_RUNTIME = "io.micronaut.function.aws.runtime.MicronautLambdaRuntime";
```
graemerocher, line 35: Is it really necessary to introduce a whole new class for a single constant?

docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
```diff
@@ -462,11 +464,8 @@ private CopyFileInstruction toCopyResourceDirectoryInstruction(java.io.File reso
             options.getBuildArgs().add("--report-unsupported-elements-at-runtime");
         } else if (buildStrategy == DockerBuildStrategy.LAMBDA) {
             JavaApplication javaApplication = getProject().getExtensions().getByType(JavaApplication.class);
-            if (!javaApplication.getMainClass().isPresent()) {
-                options.getMainClass().set("io.micronaut.function.aws.runtime.MicronautLambdaRuntime");
-            }
-            if (!options.getMainClass().isPresent()) {
```
sdelamo, line 468: @melix should not this be if/else if instead of or. Maybe this is the cause of the issue we saw.

melix: No it shouldn't. But I found some plugins which override each other's defaults, which is the real issue. This thing is spaghetti code...

graemerocher (approved):

docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
```diff
@@ -52,11 +52,17 @@
  */
 public abstract class NativeImageDockerfile extends Dockerfile implements DockerBuildOptions {
 
+    /**
+     * Main class for Provided runtime and Application Type: Micronaut Application.
+     */
+    public static final String MICRONAUT_LAMBDA_RUNTIME = "io.micronaut.function.aws.runtime.MicronautLambdaRuntime";
```
melix, line 58: I _think_ it would make more sense to have this in `MicronautDockerPlugin` or `DockerBuildStrategy`: this is not directly related to docker _files_, wdyt?

sdelamo:
close via https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/415

PR #415, opened 2022-03-25
Rework how the default runtime class is set for Lambda

melix:
Before this change, the class used to generate native lambdas was
always set to `MicronautLambdaRuntime`, independently of the fact
that the user declared a `mainClass` or not. With this change, the
main class defined by the user will _always_ be used. Only if no
main class is set, we will default to `MicronautLambdaRuntime`.

This will require changes to Micronaut Starter.


sdelamo:
The motivation of this chnage is that [we found scenarios](https://github.com/sdelamo/micronaut-cdk-workshop-gradle) where it was impossible to set the main class to anything other `MicronautLambdaRuntime`. Thus, making impossible to generate native images for functions and deploy them to a provided runtime.

sdelamo (approved):

Issue #421, opened 2022-03-28
Creating a `Function` app with lambda and graalvm adds `io.micronaut.aws:micronaut-function-aws-api-proxy` dependency.


sdelamo:
## Steps to reproduce: 

[Create a Micronaut `Function Appliation for Serverless` application in Lambda with `graalvm` and `aws-lambda`](type=FUNCTION&name=demo&package=com.example&javaVersion=JDK_11&lang=JAVA&build=GRADLE&test=JUNIT&features=graalvm&features=aws-lambda&version=3.4.0). 

The gradle plugin adds unnecessary dependencies. Which you can see by running:

Run  `./gradlew dependencies | grep proxy`

You will see: `io.micronaut.aws:micronaut-function-aws-api-proxy` and `io.micronaut.aws:micronaut-function-aws-api-proxy-test` dependencies. Those dependencies should only be added if the application type is `Micronaut Application`, the default type.

melix:
Should this be in `micronaut-launch` instead?

sdelamo:
Personally, I think it should be back then we moved a lot of logic to the plugins to hide it from users. But maybe we are being to clever.

melix:
I don't understand: the template adds those dependencies. That's what I just tried.

melix:
> Back then we moved a lot of logic to the plugins to hide it from users

This makes sense to me: distinct plugin types model different kind of applications. It's easier to fix at the plugin side: simply by upgrading, you get the right set of dependencies and you don't have to fix your build scripts. The problem here seems to be that starter itself adds dependencies:

![image](https://user-images.githubusercontent.com/316357/160369575-2a2b1784-0704-4716-b1fd-e23edd0382f0.png)

sdelamo:
I am not sure. Right now the plugin leaves you out of control and currently adds wrong dependencies. 

I think we should have  clear set of dependencies that should be added for each type and the decide where to fix it.

melix:
I'm sure we can fix the plugin if it does something wrong. But right now, do you agree that Micronaut Launch _is_ adding those dependencies, as shown in the screenshot? From what you're saying it seems that _both_ would do that (Launch and the plugin).

PR #424, opened 2022-03-28
Adds a applicationType configuration, lambda_java runtime

sdelamo:
It adds a new environment `LAMBDA_JAVA ` and `LAMBDA`. When you deploy a Micronaut function to AWS Lambda you have to choose either: 

Custom runtime or Java

Those will map s: 

`Custom Runtime` => LAMBDA
`Java` => LAMBDA_JAVA. 

It fixes: https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/421

---

Morever, I have confirmed this solves the issue mentioned here:  https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/415 

Cloning this branch
`https://github.com/sdelamo/micronaut-cdk-workshop-gradle` 

and running:
`export TEST_SUITE=GRADLE_FUNCTION_NATIVE;./release.sh` 

uses the right `-H:Class`

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -468,7 +468,12 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
         } else if (buildStrategy == DockerBuildStrategy.LAMBDA) {
             JavaApplication javaApplication = getProject().getExtensions().getByType(JavaApplication.class);
             if (applicationType == ApplicationType.DEFAULT) {
-                options.getMainClass().set(DEFAULT_LAMBDA_RUNTIME_CLASS);
+                if (!javaApplication.getMainClass().isPresent()) {
+                    options.getMainClass().set(DEFAULT_LAMBDA_RUNTIME_CLASS);
+                }
+                if (!options.getMainClass().isPresent()) {
+                    options.getMainClass().set(DEFAULT_LAMBDA_RUNTIME_CLASS);
+                }
             } else {
                 if (!options.getMainClass().isPresent()) {
                     options.getMainClass().set(javaApplication.getMainClass().orElse(DEFAULT_LAMBDA_RUNTIME_CLASS));
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -1,6 +1,7 @@
 package io.micronaut.gradle.docker;
 
 import com.bmuschko.gradle.docker.tasks.image.Dockerfile;
+import io.micronaut.gradle.ApplicationType;
 import org.graalvm.buildtools.gradle.NativeImagePlugin;
 import org.graalvm.buildtools.gradle.dsl.AgentConfiguration;
 import org.graalvm.buildtools.gradle.dsl.NativeImageOptions;
@@ -355,7 +356,7 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
     }
 
     // Everything done in this method MUST be lazy, so use providers as much as possible
-    private void setupInstructions(List<Instruction> additionalInstructions) {
+    private void setupInstructions(List<Instruction> additionalInstructions, ApplicationType applicationType) {
         DockerBuildStrategy buildStrategy = getBuildStrategy().get();
         BaseImageForBuildStrategyResolver imageResolver = new BaseImageForBuildStrategyResolver(buildStrategy);
         Provider<From> baseImageProvider = getProviders().provider(() -> new From(imageResolver.get()));
@@ -389,7 +390,7 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
                         .map(this::toCopyResourceDirectoryInstruction)
                         .collect(Collectors.toList())
         ));
-        runCommand(getProviders().provider(() -> String.join(" ", buildActualCommandLine(executable, buildStrategy, imageResolver))));
+        runCommand(getProviders().provider(() -> String.join(" ", buildActualCommandLine(executable, buildStrategy, imageResolver, applicationType))));
         switch (buildStrategy) {
             case ORACLE_FUNCTION:
                 from(new From("fnproject/fn-java-fdk:" + getProjectFnVersion()).withStage("fnfdk"));
@@ -457,7 +458,8 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
 
     protected List<String> buildActualCommandLine(Provider<String> executable,
                                                   DockerBuildStrategy buildStrategy,
-                                                  BaseImageForBuildStrategyResolver imageResolver) {
+                                                  BaseImageForBuildStrategyResolver imageResolver,
+                                                  ApplicationType applicationType) {
         NativeImageOptions options = newNativeImageOptions("actualDockerOptions");
         prepareNativeImageOptions(options);
         if (buildStrategy == DockerBuildStrategy.ORACLE_FUNCTION) {
@@ -465,8 +467,12 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
             options.getBuildArgs().add("--report-unsupported-elements-at-runtime");
         } else if (buildStrategy == DockerBuildStrategy.LAMBDA) {
             JavaApplication javaApplication = getProject().getExtensions().getByType(JavaApplication.class);
-            if (!options.getMainClass().isPresent()) {
-                options.getMainClass().set(javaApplication.getMainClass().orElse(DEFAULT_LAMBDA_RUNTIME_CLASS));
+            if (applicationType == ApplicationType.DEFAULT) {
+                options.getMainClass().set(DEFAULT_LAMBDA_RUNTIME_CLASS);
+            } else {
+                if (!options.getMainClass().isPresent()) {
+                    options.getMainClass().set(javaApplication.getMainClass().orElse(DEFAULT_LAMBDA_RUNTIME_CLASS));
+                }
             }
         }
         List<String> commandLine = new ArrayList<>();
@@ -552,11 +558,11 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
     /**
      * This is executed post project evaluation
      */
-    void setupNativeImageTaskPostEvaluate() {
+    void setupNativeImageTaskPostEvaluate(ApplicationType applicationType) {
         List<Instruction> additionalInstructions = new ArrayList<>(getInstructions().get());
         // Reset the instructions to empty
         getInstructions().set(new ArrayList<>());
-        setupInstructions(additionalInstructions);
+        setupInstructions(additionalInstructions, applicationType);
     }
 
     /**
--- a/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautExtension.java
+++ b/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautExtension.java
@@ -22,6 +22,7 @@ public abstract class MicronautExtension implements ExtensionAware {
     private final AnnotationProcessing processing;
     private final Property<String> version;
     private final Property<Boolean> enableNativeImage;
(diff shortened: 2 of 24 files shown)


docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
```diff
@@ -457,16 +458,21 @@ private CopyFileInstruction toCopyResourceDirectoryInstruction(java.io.File reso
...
         NativeImageOptions options = newNativeImageOptions("actualDockerOptions");
         prepareNativeImageOptions(options);
         if (buildStrategy == DockerBuildStrategy.ORACLE_FUNCTION) {
             options.getMainClass().set("com.fnproject.fn.runtime.EntryPoint");
             options.getBuildArgs().add("--report-unsupported-elements-at-runtime");
         } else if (buildStrategy == DockerBuildStrategy.LAMBDA) {
             JavaApplication javaApplication = getProject().getExtensions().getByType(JavaApplication.class);
-            if (!options.getMainClass().isPresent()) {
-                options.getMainClass().set(javaApplication.getMainClass().orElse(DEFAULT_LAMBDA_RUNTIME_CLASS));
+            if (applicationType == ApplicationType.DEFAULT) {
```
melix, line 470: :x: I am not sure about this change: this is going to overwrite whatever is configured for `options.getMainClass()`, which we don't want.

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautExtension.java
```diff
@@ -107,6 +117,32 @@ public MicronautExtension runtime(MicronautRuntime micronautRuntime) {
         return this;
     }
 
+    /**
+     * Configures the application type.
+     *
+     * @param applicationType The micronaut application type
+     * @return This extension
+     */
+    public MicronautExtension applicationType(String applicationType) {
```
melix, line 126: I would personally avoid those helper methods, which are not Gradle idiomatic.

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
```diff
@@ -212,6 +216,10 @@ private void configureGoogleCloudFunctionRuntime(Project project, Project p, Dep
         });
 
         // Google Cloud Function requires shadow packaging
-        p.getPluginManager().apply(ShadowPluginSupport.SHADOW_PLUGIN);
+        try {
+            p.getPluginManager().apply(ShadowPluginSupport.SHADOW_PLUGIN);
```
melix, line 220: :x: Why this try catch? Looks rather nasty. If we have to do this, then something needs to be fixed.

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
```diff
@@ -0,0 +1,155 @@
...
+    private static final String DEVELOPMENT_ONLY = "developmentOnly";
+
+    private MicronautRuntimeDependencies() {
+    }
+
+    /**
+     * A map of dependencies and scopes
+     * @return The dependencies and scopes
+     */
+    public static Map<String, List<String>> getDependencies(MicronautRuntime runtime, ApplicationType applicationType) {
```
melix, line 69: :x: I am not fan of "getter like" methods which are not getters. I'd prefer we call this `findDependenciesFor` or something similar.

minimal-plugin/src/main/java/io/micronaut/gradle/PluginsHelper.java
```diff
@@ -126,4 +126,19 @@ public static MicronautRuntime resolveRuntime(Project p) {
         }
         return micronautRuntime;
     }
+
+    public static ApplicationType resolveApplicationType(Project p) {
+        MicronautExtension ext = p.getExtensions().findByType(MicronautExtension.class);
+        Object o = p.findProperty("micronaut.applicationType");
```
melix, line 132: :x: this is not how it should be done. Instead, the convention value for `applicationType` should be set to `providers.gradleProperty("micronaut.applicationType")`

minimal-plugin/src/testFixtures/groovy/io/micronaut/gradle/AbstractGradleBuildSpec.groovy
```diff
@@ -17,6 +19,13 @@ abstract class AbstractGradleBuildSpec extends Specification {
         return GraalUtil.isGraalJVM() || System.getenv("GRAALVM_HOME")
     }
 
+    boolean containsDependency(String mavenCoordinate, String configuration) {
+        BuildResult result = build('dependencies')//  "--configuration ${configuration}"
```
melix, line 23: :x: Looks like the `configuration` parameter is unused

melix (changes requested):
I am not super comfortable with this change. IMO it's too risky for a bugfix release.

docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
sdelamo: I researched more. The way it was designed was. 

# Type: Application.

For applications of type `Application`. Applications you use `@Controller` with.  

We want to use a `mainClass` for development and test and different `mainClass` for NativeImage. 

We want users to be able during development to able to run with `./gradlew run`. 

That it is way the generated apps have a regular `Application` class and we set it as `mainClass`. That it is why we add [`function-aws-api-proxy`](https://github.com/micronaut-projects/micronaut-aws/tree/master/function-aws-api-proxy) dependency as `developmentOnly` and also in the test classpath. 

When we package as a Native image we need to use `DEFAULT_LAMBDA_RUNTIME_CLASS`. 

# For Type Function 

We want to use whather the user set in the `application` plugin configuration. We don't want to override it. 



Actually, the w. For During development,  what we want to allow users to run their app with `./gradlew run` using a regular `Application.class`. 

Only use ``

We want users to be able to use the undocumented module `function-a

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautExtension.java
sdelamo: Please, note that this methods exists for `runtime` as well. That it is why I created it. Should I remove both?

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
sdelamo: It is necessary because I GCP requires the shadowJAR plugin. 


These entries in the `where` table: 
```
 'google_function'      | ...
 'google_function'      | ...
 'google_function'      | ...
```

cause that exception to be trigger. 

However, I don't think I can apply shadow Jar gradle plugin in this test. 

I think I could if move he test to `functional-test`

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/424/commits/ab1ccf102b25155c9ad6ee9c24a47d30b0ebe0c9

minimal-plugin/src/main/java/io/micronaut/gradle/PluginsHelper.java
sdelamo: Please, note that this method is identical to `PluginsHelper::resolveRuntime`. How can I obtain a `providers` instance?

minimal-plugin/src/testFixtures/groovy/io/micronaut/gradle/AbstractGradleBuildSpec.groovy
sdelamo: I was hoping you could help me out. `AbstractGradleBulidSpec` throws `> Unknown command-line option '--configuration compileClasspath'.` if I uncomment it.  I was expecting to be able to use `"dependencies", "--configuration compileClasspath"` as a parameter.

minimal-plugin/src/main/java/io/micronaut/gradle/PluginsHelper.java
melix: Yes, but I think new stuff should be implemented properly. Most of the plugin is quite old and uses outdated patterns. You can get the `Providers` instance from the `Project#getProviders()`

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
melix: Right, so yes, if the plugin requires 3rd party the test should be moved to functional tests project. It's a bit cumbersome but it's actually the point of separating out: it allows catching errors which weren't found before (dependencies accidentally on classpath)

docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
melix: But if you remove the `if (!options.getMainClass()` call, then any change to `graalvmNative { binaries { mainClass = ... } }` will be ignored. This would be inconsistent, no?

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautExtension.java
sdelamo: Deleted in this commit https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/424/commits/9ff5488843f9fc4abeb8dd21ed33cd42f52adccf

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
sdelamo: I have moved to functional

minimal-plugin/src/main/java/io/micronaut/gradle/PluginsHelper.java
sdelamo: Changed it here: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/424/commits/8887653238e5ad6149577779673d09aff827db35

Let me know if it matches your expectations.

docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
sdelamo: I have changed it so that it is not breaking with 3.3.1 for DEFAULT applications: 

In 3.3.1 

```
} else if (buildStrategy == DockerBuildStrategy.LAMBDA) {
    JavaApplication javaApplication = getProject().getExtensions().getByType(JavaApplication.class);
    if (!javaApplication.getMainClass().isPresent()) {
        options.getMainClass().set("io.micronaut.function.aws.runtime.MicronautLambdaRuntime");
    }
    if (!options.getMainClass().isPresent()) {
        options.getMainClass().set("io.micronaut.function.aws.runtime.MicronautLambdaRuntime");
    }
}
```

With this change: 

```
} else if (buildStrategy == DockerBuildStrategy.LAMBDA) {
    JavaApplication javaApplication = getProject().getExtensions().getByType(JavaApplication.class);
    if (applicationType == ApplicationType.DEFAULT) {
        if (!javaApplication.getMainClass().isPresent()) {
            options.getMainClass().set("io.micronaut.function.aws.runtime.MicronautLambdaRuntime");
        }
        if (!options.getMainClass().isPresent()) {
           options.getMainClass().set("io.micronaut.function.aws.runtime.MicronautLambdaRuntime");
        }
    } else {
        if (!options.getMainClass().isPresent()) {
            options.getMainClass().set(javaApplication.getMainClass().orElse(DEFAULT_LAMBDA_RUNTIME_CLASS));
        }
    }
}
```

sdelamo:
Closed via https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/434 and https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/435

Issue #433, opened 2022-04-04
Micronaut Application deployed to Java Lambda Runtime adds unnecessary dependency micronaut-function-aws-custom-runtime" 


sdelamo:
### Issue description

If you select `lambda` feature and type `Application`, we automatically add `micronaut-function-aws-custom-runtime` dependency. However, it is necessary only if you deploy to a provided runtime. It is not necessary if you deploy to Java runtime.

PR #434, opened 2022-04-04
fix: Adds LAMBDA_JAVA and LAMBDA_PROVIDED runtimes

sdelamo:
Fixes: https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/433

Adds two Micronaut runtime environments:

- LAMBDA_JAVA
- LAMBDA_PROVIDED

This matches the Lambda Runtime names:

https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-lambda.Runtime.html

```
static JAVA_11
static JAVA_8
static JAVA_8_CORRETTO
static PROVIDED
static PROVIDED_AL2
```

This allows to differentiate easily the dependencies to include automatically.

### DIFF
--- a/minimal-plugin/src/main/java/io/micronaut/gradle/Dependencies.java
+++ b/minimal-plugin/src/main/java/io/micronaut/gradle/Dependencies.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2017-2022 original authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.gradle;
+
+import org.gradle.api.plugins.JavaPlugin;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Builder class to model build dependencies.
+ * @author Sergio del Amo
+ * @since 3.4.0
+ */
+public class Dependencies {
+    private static final String DEVELOPMENT_ONLY = "developmentOnly";
+
+    private final List<String> developmentOnlyDependencies;
+    private final List<String> implementationDependencies;
+    private final List<String> testImplementationDependencies;
+    private final List<String> runtimeOnlyDependencies;
+    private final List<String> compileOnlyDependencies;
+
+    private Dependencies(List<String> developmentOnlyDependencies,
+                        List<String> implementationDependencies,
+                        List<String> testImplementationDependencies,
+                         List<String> runtimeOnlyDependencies,
+                         List<String> compileOnlyDependencies) {
+        this.developmentOnlyDependencies = developmentOnlyDependencies;
+        this.implementationDependencies = implementationDependencies;
+        this.testImplementationDependencies = testImplementationDependencies;
+        this.runtimeOnlyDependencies = runtimeOnlyDependencies;
+        this.compileOnlyDependencies = compileOnlyDependencies;
+    }
+
+    public Map<String, List<String>> toMap() {
+        Map<String, List<String>> m = new HashMap<>();
+        if (isNotEmpty(developmentOnlyDependencies)) {
+            m.put(DEVELOPMENT_ONLY, developmentOnlyDependencies);
+        }
+        if (isNotEmpty(implementationDependencies)) {
+            m.put(JavaPlugin.IMPLEMENTATION_CONFIGURATION_NAME, implementationDependencies);
+        }
+        if (isNotEmpty(testImplementationDependencies)) {
+            m.put(JavaPlugin.TEST_IMPLEMENTATION_CONFIGURATION_NAME, testImplementationDependencies);
+        }
+        if (isNotEmpty(runtimeOnlyDependencies)) {
+            m.put(JavaPlugin.RUNTIME_ONLY_CONFIGURATION_NAME, runtimeOnlyDependencies);
+        }
+        if (isNotEmpty(compileOnlyDependencies)) {
+            m.put(JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME, compileOnlyDependencies);
+        }
+        return m;
+    }
+
+    private static boolean isNotEmpty(Collection<String> col) {
+        return col != null && !col.isEmpty();
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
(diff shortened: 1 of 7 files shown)


sdelamo:
@alvarosanchez I guess we should support this new runtime in the Maven plugin as well.

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntime.java
```diff
@@ -22,98 +16,69 @@
...
-
     /**
      * @return The docker build strategy
      */
     public DockerBuildStrategy getBuildStrategy() {
         return buildStrategy;
     }
-}
+
+    public static boolean isLambdaProvided(MicronautRuntime runtime) {
```
melix, line 80: :o: Why not having an instance method here? `mr.isLambdaProvided()`?

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private MicronautRuntimeDependencies() {
+    }
+
+    /**
+     * @param runtime Micronaut runtime
+     * @return The dependencies and scopes
+     */
+    public static Dependencies findApplicationPluginDependenciesByRuntime(MicronautRuntime runtime) {
+        if (runtime == LAMBDA_JAVA || isLambdaProvided(runtime)) {
```
melix, line 63: :o: Maybe add a `isLambda` method to `MicronautRuntime`.

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
```diff
@@ -0,0 +1,136 @@
...
+
+        } else if (runtime == TOMCAT) {
+            return Dependencies.builder().implementation(micronautServletDependency(ARTIFACT_ID_MICRONAUT_SERVLET_TOMCAT)).build();
+
+        } else if (runtime == JETTY) {
+            return Dependencies.builder().implementation(micronautServletDependency(ARTIFACT_ID_MICRONAUT_SERVLET_JETTY)).build();
+
+        } else if (runtime == UNDERTOW) {
+            return Dependencies.builder().implementation(micronautServletDependency(ARTIFACT_ID_MICRONAUT_SERVLET_UNDERTOW)).build();
+        }
```
melix, line 108: :o: I would add an `else if (runtime != NONE)` and throw here so that we know if we miss a branch in the future

alvarosanchez:
Agreed. If they get added here, they need to be added there too.

melix (approved):

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntime.java
sdelamo: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/434/commits/12f3d78b8a09b02732709d5e42c71bfbe68b7572

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
sdelamo: is runtime required? is NONE consider an invalid value?

melix: `NONE` is a valid value, it's a runtime which is "inject only". So we should check if it's _not_ none, in which case it means we have a missing branch.

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
sdelamo: do you want me to replace `runtime == LAMBDA_JAVA` with `runtime.isLambdaJava()`?

_MicronautRuntime_
```

    public boolean isLambdaJava() {
        return this == LAMBDA_JAVA;
    }
```

Please, note that I have added an `isLambdaProvided()` to encapsulate the deprecation of `LAMBDA` runtime .

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
sdelamo: Resolved via: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/434/commits/00132ebfc90ed068282cab06c90baa0fdd325f5f

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
melix: No, I want a method which says "the runtime is a lambda variant" which encapsulates all possible lambda cases.

minimal-plugin/src/main/java/io/micronaut/gradle/Dependencies.java
```diff
@@ -0,0 +1,120 @@
...
+                         List<String> compileOnlyDependencies) {
+        this.developmentOnlyDependencies = developmentOnlyDependencies;
+        this.implementationDependencies = implementationDependencies;
+        this.testImplementationDependencies = testImplementationDependencies;
+        this.runtimeOnlyDependencies = runtimeOnlyDependencies;
+        this.compileOnlyDependencies = compileOnlyDependencies;
+    }
+
+    public Map<String, List<String>> toMap() {
+        Map<String, List<String>> m = new HashMap<>();
```
melix, line 53: :x: Actually this should probably be a `LinkedHashMap`. Otherwise we are order sensitive, which, in general, won't be an issue, but by experience, this can cause some difficult to diagnose cache misses, in case dependencies are reordered.

melix (changes requested):
Just a minor change requested, and you'll probably have to make Sonar happy :)

minimal-plugin/src/main/java/io/micronaut/gradle/Dependencies.java
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/434/commits/5d48c7869ae3731f401da21736899196fee8921f

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautRuntimeDependencies.java
sdelamo: Resolved via https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/434/commits/62a7d53a58e03383ae79e07c725c68e49d48050d

sdelamo:
Sonar fails telling to remove the use of LAMBDA Deprecated but I think we should keep it until next mayor version.

melix:
Right you should be able to resolve it as "won't fix" by clicking on the link and logging into sonar.

sdelamo:
> Right you should be able to resolve it as "won't fix" by clicking on the link and logging into sonar.

thanks, I have done that. 

Should we create a 3.4.x branch and merge this PR?. Do you want any additional changes?

melix:
I have created a 3.3.x branch, master is now `3.4.x`.

PR #447, opened 2022-04-28
Upgrade to Micronaut AOT 1.1.0

melix:
This commit bumps Micronaut AOT to version 1.1.0, which adds a new
optimization for Netty. This optimization is disabled by default like
the others (`starter` can enable it). Because it's likely that they
will be more Netty optimizations in the future, there's a `netty`
configuration block within the `aot` one.

### DIFF
--- a/aot-plugin/src/main/java/io/micronaut/gradle/aot/AOTOptimizations.java
+++ b/aot-plugin/src/main/java/io/micronaut/gradle/aot/AOTOptimizations.java
@@ -15,10 +15,13 @@
  */
 package io.micronaut.gradle.aot;
 
+import org.gradle.api.Action;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.MapProperty;
 import org.gradle.api.provider.Property;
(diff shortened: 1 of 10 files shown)


alvarosanchez (approved):

melix:
How can I tell Sonar that coverage is done via integration testing it can't pick up?

alvarosanchez:
It reads coverage via JaCoCo, which currently it's only configured for [unit tests](https://github.com/micronaut-projects/micronaut-build/blob/master/src/main/java/io/micronaut/build/MicronautQualityReportingAggregatorPlugin.java#L67)

alvarosanchez:
Yes, you can always ignore the failure, the Sonar check is optional.

However, if you could change `micronaut-build` to aggregate coverage from integration tests too, it'd be great.

melix:
> However, if you could change micronaut-build to aggregate coverage from integration tests too, it'd be great.

It would have to inject the jacoco agent to the _build under test_ and somehow aggregate results. I'm not even sure that's doable.

PR #471, opened 2022-05-25
Pull GraalVM JDK based on system architecture

timyates:
When a user builds a native lambda via docker, docker first pulls amazonlinux:latest

If they are on an M1, docker will pull the ARM version of amazonlinux, then later on
in the dockerfile, it will cURL and unzip the X86 version of GraalVM.

This then fails when it tries to install native-image

This fix allows us to specify an architecture of GraalVM to pull, which defaults to
that of the current system.

It effectively mirrors this change made in the maven plugin for the same issue:

https://github.com/micronaut-projects/micronaut-maven-plugin/pull/369

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -56,6 +56,9 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
             // keep those in descending order
             Arrays.asList(17, 11)
     );
+    private static final String ARM_ARCH = "aarch64";
+    private static final String X86_64_ARCH = "amd64";
+    private static final String GRAAL_VERSION_CONVENTION = "22.1.0";
 
     /**
      * @return The JDK version to use with native image. Defaults to the toolchain version, or the current Java version.
@@ -68,6 +71,12 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
      */
     @Input
     public abstract Property<String> getGraalVersion();
+
+    /**
+     * @return The Graal architecture
+     */
+    @Input
+    public abstract Property<String> getGraalArch();
 
     /**
      * @return the GraalVM docker image to use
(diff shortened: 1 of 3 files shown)


docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
```diff
@@ -69,6 +72,12 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
     @Input
     public abstract Property<String> getGraalVersion();
 
+    /**
+     * @return The Graal architecture
+     */
+    @Input
+    public abstract Property<String> getGraalArch();
```
sdelamo, line 79: should this be `getArch` instead of `getGraalArch`?

sdelamo (changes requested):
Thanks, this is going to be really useful. Can you add docs?

docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
melix: I think that `graalArch` makes sense indeed.

src/docs/asciidoc/index.adoc
```diff
@@ -856,6 +856,9 @@ The advantage of allowing your dependencies to be dictated by the runtime is tha
 If you are interested in deploying your Micronaut application to AWS Lambda using GraalVM you only need to set the runtime to `lambda` and execute `./gradlew buildNativeLambda`.
 This task will generate a GraalVM native image inside a Docker container and then it will create the file `build/libs/your-app.zip` file ready to be deployed to AWS Lambda using a custom runtime. See more information in {aws-docs}[Micronaut AWS documentation].
 
+The plugin will detect the host operating system architecture (based on the `os.arch` Java system property) and will install the corresponding GraalVM  binary distribution inside the Docker image.
+This means that when running packaging from an X86_64 (Intel/AMD) machine, the produced native image will be an `amd64` binary, whilst on an ARM host (such as the new Mac M1) it will be an `aarch64` binary.
```
melix, line 860: Can you add an example of how to override that?

sdelamo:
this is a new feature? This should go to 3.5.0 instead of 3.4.1?

src/docs/asciidoc/index.adoc
timyates: Example added 👍

melix (approved):

PR #478, opened 2022-06-09
Add test resources plugin

melix:
This commit introduces a new plugin, `io.micronaut.test-resources`
which integrates with Micronaut Test Resources. It provides support
for starting the test resources server in the background, and lets
the user configure test resources support.

This plugin can be used in combination with other "minimal" plugins:

```
plugins {
     id 'io.micronaut.minimal.application'
     id 'io.micronaut.test-resources'
}
```

It is automatically applied by the `application` and `library`
plugins. Support for test resources is only enabled if the target
Micronaut version is at least 3.5.

The plugin introduces a configuration block:

```
micronaut {
   ...
   testResources {
      // test resources configuration goes here
   }
}
```

By default, the plugin will perform classpath inference
to determine which test resources module to include. There
are several configuration options:

```
testResources {

    enabled = true // by default, only if Micronaut 3.5+

    version = "1.2" // override Test Resources version

    classpathInference = false // true by default

    // Add explicit Micronaut Test resources modules
    // in case inference fails to find them
    additionalModules.addAll("kafka", "hivemq")

    // Instead of using a randomized port, use an
    // explicit port for the server
    explicitPort.set(15455)
}
```

By default, the server will only accept requests from
this build, thanks to a randomly generated access token.
However, if an explicit port is used, token generation
is disabled: this is so that multiple, concurrent builds
can access the same test resources server (which may be
useful when testing interations between a client and
a server in a messaging solution for example).

Last but not least, test resources try to infer dependencies
to be added on the server classpath. There may be cases
where it fails finding dependencies it should add, in
which case you can explicitly add dependencies:

```
dependencies {
    testresources 'some.sql:driver:1.0'
}
```

The server is integrated in tests and development mode.


melix:
@alvarosanchez @graemerocher for now the PR is in draft state, so that we can discuss the default behavior. I have made it so that the `application` and `library` plugins, which are typically used in projects generated by starter, automatically apply the test resources plugin. While it may sound like a good idea, it comes with a price: all tests, or the `run` task, will require starting the server beforehand, even if your application doesn't need test resources (we can't know upfront if it will need them or not, since it's determined at runtime by resolving missing properties).

It is possible, however, to disable support altogether:

```
micronaut.testResources.enabled = false
```

Let's make sure we agree on the defaults before merging. I will add documentation in another PR (first we need to document test-resources itself).

graemerocher:
Yeah I don't think we shouldn't add it automatically to the `application` and `library` plugins. We can modify starter to add it so users who don't want it can remove it.

graemerocher:
By the way is there a way to start the test resource server asynchronously so it doesn't block starting the main application server?

dstepanov:
`additionalModules.addAll("kafka", "hivemq")`
Maybe we can have some constants? 

Should we support something like `explicitModules` if somebody wants to set it without the auto-recognition?

melix:
> Maybe we can have some constants?

Good idea!

melix:
> Should we support something like explicitModules if somebody wants to set it without the auto-recognition?

If you disable `classpathInference` and use `additionalModules` then it would have the same effect. Or is it something different?

melix:
> Yeah I don't think we shouldn't add it automatically to the application and library plugins. We can modify starter to add it so users who don't want it can remove it.

Ok I removed it, it has to be applied explicitly now.

PR #480, opened 2022-06-13
Add coverage for Gradle TestKit builds

melix:


melix:
Merged even if coverage still not showing up: there are beneficial changes in any case.

PR #485, opened 2022-06-16
Introduce support for multi-project builds

melix:
Builds on top of #484

This commit introduces a way to provide a test resources service
which is shareable between projects of a multi-project build.
Without this, each project of a multi-project build applying the
test resources plugin would start a test resources service
independently of the others, which means that if 2 projects need
the same containers types, they wouldn't be able to re-use the
same container (unless they use the `shared` mode and make sure
that each project declares all test resolvers which are required
by all projects).

With this change, it is now possible to declare a project which
is the "test resources supplier". This project is responsible for
declaring all resources it can serve. Then each project willing
to use this test resources server need to apply a _different_
plugin, `io.micronaut.test-resources-consumer` and declare what
project provides the test resources server:

```gradle
micronaut {
   testResources.targetProject = ':testresources-supplier'
}
```

The configuration will then automatically be injected, and the
test resources service lifecycle will be handled appropriately.


graemerocher (approved):

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesPlugin.java
```diff
@@ -129,6 +123,23 @@ private void configurePlugin(Project project) {
...
+    private Path createStopFile(Project project) {
+        Path stopAtEndFile;
+        try {
+            File asFile = project.getLayout().getBuildDirectory().file("test-resources/" + UUID.randomUUID()).get().getAsFile();
+            if (asFile.getParentFile().isDirectory() || asFile.getParentFile().mkdirs()) {
+                asFile.deleteOnExit();
+                stopAtEndFile = asFile.toPath();
+                Files.deleteIfExists(stopAtEndFile);
+            } else {
+                throw new IOException("Could not create directory for test resources stop file");
```
vlsi, line 135: Would you please mention the problematic filename in the exception message?

samples/test-resources/multiproject/app1/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+}
+
+micronaut {
+    runtime("netty")
+    testRuntime("junit5")
+    processing {
+        incremental(true)
+        annotations("demo.*")
+    }
+    testResources.targetProject = ':testresources'
```
vlsi, line 34: Frankly speaking, I never used Micronaut before, and everything in the build file looks understandable except `testResources.targetProject = ':testresources'`

It looks like there are three entities "testResources", "targetProject", and ":testresources", and none of them makes sense to me at this point 🤷‍♂️

vlsi: Have you considered regular `dependencies` block instead?

For instance:

```groovy
dependencies {
  testResources(project(":testresources"))
}
```

I might be wrong, however, it does clarify the intention: "we depend on test resources provided by `:testresources` project). It would be compatible with project accessors.

melix: I have a doco PR coming which should clarify this.

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesPlugin.java
melix: will do

PR #486, opened 2022-06-17
Add test resources documentation

melix:
@vlsi hope this clarifies what you pointed out in the multi-project support PR


vlsi:
@melix ,

I think `targetProject` is not documented.
However, I would prefer something that can be understood by a mere looking at the build file without searching the documentation for each and every property.

melix:
> However, I would prefer something that can be understood by a mere looking at the build file without searching the documentation for each and every property.

What would you suggest? We need a way to tell which project is the test resources provider. We can't use a `Configuration`, because it doesn't make sense: we don't want a list of dependencies, we want a project reference. We don't want a `Project` instance either, because this is capturing too much state.

vlsi:
> We can't use a Configuration, because it doesn't make sense: we don't want a list of dependencies, we want a project reference

Well, I thought configuration+dependency would be easier to understand and extend.

As far as I see you use "project reference" only for creating a dependency: https://github.com/micronaut-projects/micronaut-gradle-plugin/search?q=gettargetproject, so using the dependency looked reasonable to me.

At the same time, a regular dependency would support typesafe project accessors while `targetProject=...` would not support project accessors :-/

>I think targetProject is not documented.
> It is.

I think the only place where `targetProject` appears is the case when "targetProject" is used in the sample build plan.

https://github.com/search?q=org%3Amicronaut-projects+targetProject&type=code

At the same time, I do not see `targetProject` being added by the current PR 🤷‍♂️ :
targetProject usages in PR

Would you please highlight where `targetProject` is mentioned?

melix:
> Would you please highlight where targetProject is mentioned?

Precisely in the screenshot you pasted, with the callout :)

> Well, I thought configuration+dependency would be easier to understand and extend (e.g. it would support adding dependency on several testresources projects).

Right, I think dependency on multiple test resource projects would sound a bit scary, but I can see the appeal.

vlsi:
>Precisely in the screenshot you pasted, with the callout :)

Now I see. The name of "target" looks weird, as it feels something like "deploy test resources to a target project X"

melix:
@vlsi what about [1ffaf9a](https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/486/commits/1ffaf9a271a2aabba09f99db08bf87a5c38b3259)?

src/docs/asciidoc/index.adoc
```diff
@@ -1182,6 +1191,272 @@ The 2 tasks which are available for this are:
...
+
+[source, groovy, subs="verbatim,attributes", role="multi-language-sample"]
+----
+plugins {
+    id "io.micronaut.application" version "{micronaut-version}"
+    id "io.micronaut.test-resources-consumer" version "{micronaut-version}" // <1>
+}
+
+micronaut {
+    testResources.targetProject = ':shared-testresources'           // <2>
```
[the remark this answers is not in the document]
melix: good catch, fixed

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesConsumerPlugin.java
```diff
@@ -39,23 +37,36 @@ public class MicronautTestResourcesConsumerPlugin implements Plugin<Project> {
...
+
+    private Configuration createTestResourcesExtension(Project project) {
+        return project.getConfigurations().create(MicronautTestResourcesPlugin.TESTRESOURCES_CONFIGURATION, conf -> {
+            conf.setCanBeConsumed(false);
+            conf.setCanBeResolved(false);
+            conf.setDescription("Used to declare projects which provide test resources");
+            conf.getDependencies().whenObjectAdded(dep -> {
+                if (dep instanceof ProjectDependency) {
+                    ProjectDependency projectDependency = (ProjectDependency) dep;
+                    projectDependency.attributes(attrs -> configureUsageAttribute(project, attrs));
```
vlsi, line 54: I thought the attributes should be configured at the configuration level

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesConsumerPlugin.java
```diff
@@ -39,23 +37,36 @@ public class MicronautTestResourcesConsumerPlugin implements Plugin<Project> {
     public void apply(Project project) {
         PluginManager pluginManager = project.getPluginManager();
         pluginManager.apply(MicronautBasePlugin.class);
-        TestResourcesConsumerConfiguration testResourcesExtension = createTestResourcesExtension(project);
-        pluginManager.withPlugin("io.micronaut.component", unused -> {
-            project.afterEvaluate(p -> p.getConfigurations().all(cnf -> configureDependencies(p, testResourcesExtension, project.getDependencies(), cnf)));
+        Configuration testResourcesConfiguration = createTestResourcesExtension(project);
+        pluginManager.withPlugin("io.micronaut.component", unused ->
+                project.afterEvaluate(p -> p.getConfigurations().all(cnf -> configureDependencies(testResourcesConfiguration, cnf)))
```
vlsi, line 42: Is this afterEvaluate still needed?

vlsi:
@melix, thank you. I think it is indeed better. However, now I'm feeling guilty for the disruption :)

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesConsumerPlugin.java
```diff
@@ -39,23 +37,36 @@ public class MicronautTestResourcesConsumerPlugin implements Plugin<Project> {
...
+        );
+    }
+
+    private Configuration createTestResourcesExtension(Project project) {
+        return project.getConfigurations().create(MicronautTestResourcesPlugin.TESTRESOURCES_CONFIGURATION, conf -> {
+            conf.setCanBeConsumed(false);
+            conf.setCanBeResolved(false);
+            conf.setDescription("Used to declare projects which provide test resources");
```
vlsi, line 50: "test resources" -> "micronaut test resources"?

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesConsumerPlugin.java
melix: They could, for configurations which can be resolved. However if we put it on the configuration, then _all_ dependencies will be resolved with those attributes. This is not what we want here, since only the test resources dependencies need to resolve the test resources variant.

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesConsumerPlugin.java
melix: Technically speaking it's not specific to Micronaut: the same server could be used in a Spring app, if they implemented a resolution mechanism.

test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesConsumerPlugin.java
melix: I will double check. A `configureEach` should work in theory. In the other plugin I had to defer configuration.

vlsi: I thought Configurations were never lazy, so .all should be fine

melix: They are not, but if your `.all` action executes synchronously if the configuration exists, and asynchronously if not. This caused some weird bugs.

PR #491, opened 2022-06-23
Fix test resources dependencies at runtime

melix:
Note: only the last commit is relevant in this PR, since it's built on top of unmerged PR.

In case a test resource written in `testResources` is using extra
dependencies, then those dependencies were added to the compile
classpath of the test resources, but not in the server runtime
classpath, which lead to class not found errors and impossibility
to load the test resource resolver.

We discovered the problem while pairing on a custom resource resolver with @timyates .


timyates (approved):
Such a simple change when you know how 🙇

PR #500, opened 2022-07-14
Support AppCDS to improve startup time of test resource server

graemerocher:
This PR allows the server to be optimised with AppCDS 

Unfortunately due to issues with the server we have to limited the classes in the class list when creating the archive, but this doesn't seem to impact startup time.


alvarosanchez:
Most of this should really belong to https://github.com/micronaut-projects/micronaut-test-resources/tree/master/test-resources-build-tools

graemerocher:
yeah it is a draft to just get it working... Cedric will look when he is back from holiday

graemerocher:
FWIW it cuts startup time for me from 650ms to 280ms

melix:
> Most of this should really belong to https://github.com/micronaut-projects/micronaut-test-resources/tree/master/test-resources-build-tools

It's actually hard to make it part of the common code, since this belongs to the invocation logic: we're _already_ in the build tool specific code.

melix:
Superceded by #508

PR #507, opened 2022-07-20
Fix support for GraalVM 22.2

melix:
This commit adds a workaround for GraalVM 22.2 native-image builder running
on module path. Because Micronaut uses features which use internal APIs (and
no workaround for that), building on 22.2 fails. This commit adds a flag
which will let users build by adding a module export.

See https://github.com/oracle/graal/pull/4468


alvarosanchez:
You should update `.github/workflows/graalvm.yml` accordingly as per https://github.com/micronaut-projects/micronaut-project-template/pull/247

PR #524, opened 2022-08-19
Add a plugin to build CRaC enhanced docker images

timyates:
Currently only supports APPLICATION.

This plugin adds several tasks to the CRaC group.

The flow is:

1. Copy or create scripts required for checkpointing to build/docker/main/checkpoint
2. Create a Dockerfile (build/docker/main/Dockerfile.CRaCCheckpoint) for checkpointing image
    - This dockerfile installs the CRaC JDK
    - runs the app, then triggers a checkpoint
    - checkpoint files are written to build/docker/main/cr by the container via a volume
3. Runs this dockerfile in a container
4. Awaits the completion of this container (checks for failure in the logs)
5. Creates a Dockerfile for running the checkpointed app
    - This dockerfile copies the CRaC JDK from the Checkpoint image to prevent another download
    - Copies the checkpoints from build/docker/main/cr into the expected place
6. Builds an image from this dockerfile

Currently, this is only known to work on an Ubuntu 18 host, or OSX x86 machines


timyates:
Not yet ready for review, as I've not yet tested it manually on an x86 machine

crac-plugin/src/main/java/io/micronaut/gradle/crac/CRaCCheckpointDockerfile.java
```diff
@@ -0,0 +1,136 @@
...
+import org.gradle.api.tasks.TaskAction;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CRaCCheckpointDockerfile extends Dockerfile {
+    public static final String DEFAULT_WORKING_DIR = "/home/app";
+
+    @Input
+    private final Property<String> baseImage;
```
melix, line 21: :x: don't use fields, but instead make `CRaCCheckpointDockerfile` abstract and use `public abstract Property<String> getBaseImage()`. The conventions should be set in the plugin class.

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
```diff
@@ -0,0 +1,247 @@
...
+    public static final String BUILD_DOCKER_DIRECTORY = "docker/";
+
+    @Override
+    public void apply(Project project) {
+        configurePlugin(project);
+    }
+
+    private void configurePlugin(Project project) {
+        ExtensionContainer extensions = project.getExtensions();
+        MicronautExtension micronautExtension = extensions.getByType(MicronautExtension.class);
```
melix, line 47: :x: This will fail if the CrAC plugin is applied before, say, the `io.micronaut.application` plugin. If you need the `MicronautExtension` to be present, then you need to apply the `MicronautBasePlugin`. I guess this won't work either if you don't apply the `MicronautDockerPlugin`?

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
```diff
@@ -0,0 +1,247 @@
...
+        CRaCConfiguration configuration = createCRaCConfiguration(project);
+        NamedDomainObjectContainer<MicronautDockerImage> dockerImages = (NamedDomainObjectContainer<MicronautDockerImage>) micronautExtension.getExtensions().findByName("dockerImages");
+        createCheckpointDockerImage(project, dockerImages.findByName("main"), configuration);
+    }
+
+    private CRaCConfiguration createCRaCConfiguration(Project project) {
+        MicronautExtension micronautExtension = PluginsHelper.findMicronautExtension(project);
+        CRaCConfiguration crac = micronautExtension.getExtensions().create("crac", CRaCConfiguration.class);
+        crac.getEnabled().convention(true);
+        crac.getImageName().convention("Woo");
```
melix, line 57: Is this a real name?

crac-plugin/src/main/resources/warmup.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# Warm up the app.  For now, just call curl, we may want a way for this to be configurable
+curl http://localhost:8080
```
melix, line 4: This would fail if the user configures a different port for the Micronaut application, right?

timyates: Yes, thats why I allow the user to specify their own warmup script to replace this default...

I need to add the same for "readiness"

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
timyates: OOpps...

It was vestigial...  Removed 👍

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
timyates: Yeah, I couldn't decide on the best policy...or understand the inner workings...

Should this plugin just be additive to the application plugin?  ie:

```
plugins {
  id('io.micronaut.application')
  id('io.micronaut.crac')
}
```

Or should it apply the docker plugin and base plugins itself (as it needs them?)

Would that then fail if someone had:

```
plugins {
  id('io.micronaut.application')
  id('io.micronaut.docker')
  id('io.micronaut.crac')
}
```
?

melix: That's a good question. I'm leaning towards thinking that the plugin _adds a functionality_ to other plugins, similarly to the `aot` or `test-resources` plugin. Therefore, the plugin should, IMO, _react_ to the presence of the `docker` plugin. That is to say you would have:

```
plugins {
  id('io.micronaut.application')
  id('io.micronaut.crac')
}
```

(because `application` applies `docker`). Nevertheless, your plugin requires the Micronaut extension, so must apply the `MicronautBasePlugin`. Then it would be a matter of wrapping the configuration into a:

```
plugins.withPlugin("io.micronaut.docker") { ... }
```

block

crac-plugin/src/main/resources/warmup.sh
timyates: Readiness handled here https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/524/commits/2a0cb2dfb6d0c0e4a90c441467289a843552766b

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
timyates: Cool.  Do we need to guess mistakes the user is going to make?

ie:

```
        PluginManager pluginManager = project.getPluginManager();
        pluginManager.apply(MicronautBasePlugin.class);
        if (!pluginManager.hasPlugin("io.micronaut.docker")) {
            throw new GradleException("The Micronaut CRaC plugin requires the Micronaut Docker plugin to be enabled");
        }
```

melix: No, you must not use `hasPlugin` (this method is eager, so you don't want to do this). You should use `pluginManager.withPlugin(...)` instead, which will execute _only_ when the docker plugin is applied.

timyates: Urgh...  I wish Greedy methods were annotated as such

Cool, how's https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/524/commits/14766a6a4572b6968dfd28a6ba915e5ba3b4b5c6 look?

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
```diff
@@ -49,8 +48,9 @@ public class MicronautCRaCPlugin implements Plugin<Project> {
     public void apply(Project project) {
         PluginManager pluginManager = project.getPluginManager();
         pluginManager.apply(MicronautBasePlugin.class);
-        pluginManager.apply(MicronautDockerPlugin.class);
-        configurePlugin(project);
+        pluginManager.withPlugin("io.micronaut.docker", plugin -> {
+            configurePlugin(project);
```
melix, line 52: :+1:

crac-plugin/src/test/groovy/io/micronaut/gradle/crac/BaseCracGradleBuildSpec.groovy
```diff
@@ -40,8 +40,7 @@ abstract class BaseCracGradleBuildSpec extends AbstractGradleBuildSpec {
     String getPluginsBlock() {
         """
             plugins {
-                id "io.micronaut.minimal.application"
-                id "io.micronaut.docker"
+                id "io.micronaut.application"
```
melix, line 43: I would keep the test using `minimal.application` and `docker`, possibly shuffling order so that you check that it works when `crac` comes first.

timyates: Good idea!  Pushed https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/524/commits/a1eb3bc720680375b2ae0098ddcfcd24b79e7c6b

timyates:
Added @alvarosanchez to see what the plugin does for Gradle

crac-plugin/src/main/java/io/micronaut/gradle/crac/CRaCCheckpointDockerfile.java
```diff
@@ -0,0 +1,153 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@CacheableTask
+public class CRaCCheckpointDockerfile extends Dockerfile {
+    public static final String DEFAULT_WORKING_DIR = "/home/app";
+
+    @Input
+    private final Property<String> baseImage;
```
melix, line 24: Still think this task should be converted to an abstract class and let Gradle implement the getters injection.

crac-plugin/src/main/java/io/micronaut/gradle/crac/CRaCCheckpointDockerfile.java
```diff
@@ -0,0 +1,153 @@
...
+    public Property<String> getTargetWorkingDirectory() {
+        return targetWorkingDirectory;
+    }
+
+    @TaskAction
+    @Override
+    @SuppressWarnings("java:S106") // System.out
+    public void create() {
+        super.create();
+        System.out.println("Dockerfile written to: " + getDestFile().get().getAsFile().getAbsolutePath());
```
melix, line 58: Maybe say `CrAC Dockerfile`?

melix (approved):
I have created a `3.5.x` branch and bumped `master` to `3.6.0-SNAPSHOT` so that you can merge this safely.

melix (approved):

melix:
If this is targeted for Micronaut 3.7, it needs to be merged into the `3.6.x` branch.

sdelamo (changes requested):
It misses documentation.

sdelamo (approved):

sdelamo:
this looks good to me.

PR #537, opened 2022-09-12
Allow to specify GraalVM releases URL

mrjoe7:
Having ability to specify GraalVM release archive is useful in case there is a need for running offilne builds.


melix (approved):

melix:
Thanks for your contribution! Would you mind rebasing your changes onto `3.6.x` so that it makes into the next release (`master` is for Micronaut 4 which is due later this year).

PR #538, opened 2022-09-14
fix(MicronautDockerPlugin): put local resources and classes at the beginning of the Class-Path

meriouma:
We're having an issue where our `application.yml` and `logback.xml`, located in the `resources` folder, are being ignored as one if the lib in the class-path already include a file with that name. I think it makes sense to put the current application's resources before the libs.

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java
@@ -163,11 +163,11 @@ public class MicronautDockerPlugin implements Plugin<Project> {
                     Configuration runtimeClasspath = project.getConfigurations()
                             .getByName(RUNTIME_CLASSPATH_CONFIGURATION_NAME);
 
+                    classpath.add("resources/");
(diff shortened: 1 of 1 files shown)


sdelamo:
> We're having an issue where our `application.yml` and `logback.xml`, located in the `resources` folder, are being ignored as one if the lib in the class-path already include a file with that name. I think it makes sense to put the current application's resources before the libs.

Please, contact the library authors. libraries should not publish `application.yml` files or `logback.xml` files.

meriouma:
> > We're having an issue where our `application.yml` and `logback.xml`, located in the `resources` folder, are being ignored as one if the lib in the class-path already include a file with that name. I think it makes sense to put the current application's resources before the libs.
> 
> Please, contact the library authors. libraries should not publish `application.yml` files or `logback.xml` files.

Regardless of why the issue appears, I still think it makes more sense to put the resources before the libraries on the classpath. I believe this is what happens when the normal jar file is being run, as more often than not, the resources are included in the application jar.

melix:
I agree that putting classes and resources upfront, compared to transitive dependencies, makes more sense.

PR #547, opened 2022-09-20
Fixes #543 APK should update with no caching enabled

### LINKED ISSUE #543: Optimisation: Docker multi-stage build for native image should not cache APK data in the final stage
### Issue description

Using version 3.5.3 of the Gradle plugins produces a `DockerfileNative` such as.

```Dockerfile
FROM ghcr.io/graalvm/native-image:ol7-java17-22.2.0 AS graalvm
WORKDIR /home/app
COPY layers/classes /home/app/classes
RUN mkdir /home/app/config-dirs
COPY config-dirs/generateResourcesConfigFile /home/app/config-dirs/generateResourcesConfigFile
RUN native-image -cp /home/app/libs/*.jar:/home/app/resources:/home/app/application.jar --no-fallback -H:Name=application -J--add-exports=org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk=ALL-UNNAMED -J--add-exports=org.graalvm.nativeimage.builder/com.oracle.svm.core.configure=ALL-UNNAMED -H:ConfigurationFileDirectories=/home/app/config-dirs/generateResourcesConfigFile -H:Class=ch.onstructive.www.Application
FROM frolvlad/alpine-glibc:alpine-3.12
RUN apk update && apk add libstdc++
COPY --from=graalvm /home/app/application /app/application
ENTRYPOINT ["/app/application"]
```

In the final stage the Dockerfile updates the `apk` index and adds `libstdc++` to the Docker image. The index isn't thrown away and remains in the image.

Instead of calling 

```sh
RUN apk update && apk add libstdc++
```
the plugin should add the `--no-cache` option, which will purge the data afterwards. (see https://stackoverflow.com/a/49119046/960875)

```sh
RUN apk --no-cache update && apk add libstdc++
```

saw303:
Small optimisation to reduce the resulting Docker image size. By adding the `--no-chache` option the package list does not get included in the Docker layer, which is fetched by `apk update.

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -463,7 +463,7 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
                 // mandatory dependency for alpine-glibc docker images
                 runCommand(getProviders().provider(() -> {
                     if (baseImageProvider.get().getImage().contains("alpine-glibc")) {
-                        return "apk update && apk add libstdc++";
+                        return "apk --no-cache update && apk add libstdc++";
                     }
                     return "";
                 }));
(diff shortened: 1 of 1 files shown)


saw303:
Sorry, I messed this PR up. Going to create a new one

PR #551, opened 2022-09-22
Let the user choose a namespace for the shared test resources service

melix:


test-resources-plugin/src/main/java/io/micronaut/gradle/testresources/MicronautTestResourcesPlugin.java
```diff
@@ -250,6 +251,7 @@ private TestResourcesConfiguration createTestResourcesConfiguration(Project proj
                             return Boolean.parseBoolean(str);
                         })
         );
+        testResources.getSharedServerNamespace().convention(providers.environmentVariable("SHARED_TEST_RESOURCES_NAMESPACE"));
```
timyates, line 254: 😻

timyates (approved):
Does this need documenting as well?

timyates (approved):

PR #557, opened 2022-09-27
Rework how the Kotlin plugin is configured

melix:


melix:
This doesn't fix the classloading problems that we may encounter in projects (because of conditional application of plugins), but it is technically more correct.

alvarosanchez (approved):

PR #564, opened 2022-10-06
Bump pid to 599 to avoid pid conflicts when restoring

richarddd:
This allows users to restore without running docker with `--privileged`

If pid is 599 there is a low likelihood will conflict with an existing one, so CRIU does not need to fork the process or use pid namespaces to restore. This means that we can run without elevated permissions

### DIFF
--- a/crac-plugin/src/main/resources/checkpoint.sh
+++ b/crac-plugin/src/main/resources/checkpoint.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+
+# Bump pid to avoid pid conflicts when restroing
+echo 599 > /proc/sys/kernel/ns_last_pid
+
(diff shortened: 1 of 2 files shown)


timyates (approved):
Nice!

We can't use `--cap-add` in the checkpointing task until the next major version of Micronaut when we have access to it via the docker plugin.

Raised https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/565 to not forget

richarddd:
> Nice!
> 
> We can't use `--cap-add` in the checkpointing task until the next major version of Micronaut when we have access to it via the docker plugin.
> 
> Raised #565 to not forget

It might be that checkpointing needs more permissions than restore but I don't remember. You would have to test and see 🤞

PR #580, opened 2022-11-03
Fix "run" task in IntelliJ

melix:
This commit adds a workaround specifically for IntelliJ, which generates synthetic `JavaExec` tasks whenever the user clicks on the run button of a `main` method. This task is _not_ the `run` task, so previously it didn't trigger the initialization of test resources. With this commit, we "detect" such synthetic tasks and fix the classpath of the task so that it includes the test resources properties.


graemerocher (approved):

MT-Jacobs:
This is an awesome find! I had no idea why test resources wouldn't always start up.

PR #582, opened 2022-11-08
Fix CRaC for ubuntu 22.04

timyates:
Previously, the CRaC JDK fails if using Ubuntu 22.04

After checking with the Azul team, this issue only affects Mac and WSL docker clients where the kernel has not yet been updated to support it.

This fix sets and environment variable that has no effect for previous versions, but fixes the kernel settings for 22.04.

I have made the default 22.04, and updated the docs accordingly


timyates:
I have added this to the 3.8.0 project board...  Trying to add it to the 3.6.4 milestone, but github isn't letting me create one

melix (approved):

PR #585, opened 2022-11-10
Allow setting a network for the crac checkpoint

timyates:
If using mysql in a container, we need to be able to specify the network it is running on.

### DIFF
--- a/crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
+++ b/crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
@@ -162,6 +162,9 @@ public class MicronautCRaCPlugin implements Plugin<Project> {
             task.targetImageId("checkpoint:latest");
             task.getContainerName().set(CRAC_CHECKPOINT);
             task.getHostConfig().getPrivileged().set(true);
+            if (configuration.getNetwork().isPresent()) {
+                task.getHostConfig().getNetwork().set(configuration.getNetwork());
+            }
             String local = project.getLayout().getBuildDirectory().dir(BUILD_DOCKER_DIRECTORY + imageName + "/cr").map(d -> d.getAsFile().getAbsolutePath()).get();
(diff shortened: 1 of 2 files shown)


sdelamo (approved):

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
```diff
@@ -162,6 +162,9 @@ private CheckpointTasksOfNote configureCheckpointDockerBuild(Project project,
             task.targetImageId("checkpoint:latest");
             task.getContainerName().set(CRAC_CHECKPOINT);
             task.getHostConfig().getPrivileged().set(true);
+            if (configuration.getNetwork().isPresent()) {
```
melix, line 165: :x: This is incorrect. You are setting a value on the task, at configuration time, when the task is registered. This means that any change to this configuration _after_ registration will fail.

Instead, you must have something like `task.getHostConfig().getNetwork().convention(configuration.getNetwork())`. I also think you'll need to mark the `network` property as `@Optional` if it's not already.

melix (changes requested):

crac-plugin/src/main/java/io/micronaut/gradle/crac/MicronautCRaCPlugin.java
timyates: One day, I'll get this right...  PTAL https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/585/commits/b673f28ee49f9f1155b333388452054696af8f7e

timyates: It looks like we don't need the `@Optional`, but I'm inclined to leave it there for documentation purposes

melix: What I meant was on `HostConfig` but it's already there, so we're good to go.

melix (approved):

PR #599, opened 2022-11-22
Upgrade to Micronaut 4

melix:
This commit upgrades the plugin to use Micronaut 4 by default. It switches from using the `io.micronaut:micronaut-bom` to `io.micronaut.bom:micronaut-bom`.

There are still test failures due to incompatibilities with Micronaut Data.


melix:
The CRaC tests are failing at this stage:

> RUN release="$(curl -sL https://api.github.com/repos/CRaC/openjdk-builds/releases/latest)"

which I think is related to the GitHub API rate limits which we are constantly hitting those days ...

graemerocher (approved):

PR #600, opened 2022-11-22
Remove `testNativeImage`

melix:
This commit removes the `testNativeImage` task which was specific to Micronaut, and executed the JVM test suite against a native image (this is different from the GraalVM NBT `nativeTest` task which compiles the test suite itself into a native binary).


graemerocher (approved):
Note that some of our own modules use `testNativeImage` 

These will need to be migrated when upgrading

PR #601, opened 2022-11-24
Add ability to debug the test resources service

melix:
This is mostly useful when debugging Micronaut Test Resources itself, but it could be useful for users developing their own test resources too.


alvarosanchez (approved):
Don't you need to add it to the documentation?

melix:
> Don't you need to add it to the documentation?

It's in the javadocs of the task type already. Documenting more globally would probably require exposing a DSL element in `micronaut { testResources { ... } }` to enable it. This PR is mostly to ease my own development process for now :)

PR #611, opened 2022-12-02
Stop container name collisions for the checkpoint container

timyates:
If the dockerBuildCrac task fails when generating a checkpoint, the checkpoint container remains in docker

This results in manual cleanup being required before a re-run can be performed

To fix this, this PR does not give the container a name (so it gets allocated a random one by docker)

This PR also names the image based on the project root name and path, which will prevent images potentially overwriting each other if two tasks are run at the same time on two separate projects


melix (approved):

timyates:
All green (after a restart of the hanging java 17 action)

PR #616, opened 2022-12-05
Add sanity check in case SnakeYAML is missing

melix:


graemerocher (approved):

graemerocher:
need the same for Maven /cc @alvarosanchez

PR #628, opened 2023-01-11
update to graal 22.3.0

sdelamo:
I think we should create a new minor 3.7.x and merge this Pr into that branch.


sdelamo:
I have updated the GraalVM release checklist so that we don't forget this in the future

https://github.com/micronaut-projects/micronaut-core/wiki/GraalVM-Release-Checklist

melix (approved):

PR #638, opened 2023-02-14
Enable GraalVM metadata repository

melix:
This commit enables use of the GraalVM metadata repository by default. It changes how the files are copied when working on Docker images, because previously, we had a single directory of resources (the generated one), but now it is possible to have multiple directories. For example, if netty is present on classpath, we will need to copy the configuration files for netty.

In practice, when we prepare the docker build context, we need to scan those directories and copy them. It is however possible that several directories have the same name, because the directory names are derived from GAV coordinates, so the same version can appear multiple times.

In order to:
   - avoid overwriting configuration files for several modules which have the same version
   - make things easier to debug when reading the docker files

we add inference to figure out if a configuration directory comes from the metadata repository, in which case we wouldn't use only the directory name, but the 3 last directories which correspond to the GAV coordinates. For example, instead of a directory named `1.0.0`, we would have a directory named `org.company:artifact:1.0.0`.

### DIFF
--- a/aot-plugin/src/main/java/io/micronaut/gradle/aot/AbstractMicronautAotCliTask.java
+++ b/aot-plugin/src/main/java/io/micronaut/gradle/aot/AbstractMicronautAotCliTask.java
@@ -34,6 +34,7 @@ import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


graemerocher:
does this work for users behind a VPN now?

melix:
> does this work for users behind a VPN now?

By default the plugin uses an [artifact published alongside the plugin](https://repo1.maven.org/maven2/org/graalvm/buildtools/graalvm-reachability-metadata/0.9.19/graalvm-reachability-metadata-0.9.19-repository.zip) which is published on Maven Central and using regular dependency resolution, so it should work behind a proxy or VPN. It may not be the latest release of metadata available, though, in which case a manual URL needs to be configured.

graemerocher (approved):

PR #656, opened 2023-03-01
Upgrade to Micronaut Test Resources 1.2.4

melix:


melix:
> You've probably answered this already but... why isn't renovate picking up this?

I guess because our workflow says "every weekend"

PR #657, opened 2023-03-02
Rework validation module annotation processor handling

graemerocher:
Currently the plugin always applies the old `micronaut-validation` module which means the plugin is currently broken with:

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':annotationProcessor'.
   > Could not find io.micronaut:micronaut-validation:.
     Required by:
         project :
```

This reworks the code so the validation processor is only applied if a validation module exists on the classpath


melix (approved):
I don't like that hardcoded group to processor map and friends, makes things hard to follow. I'll think about how we can improve this.

PR #671, opened 2023-03-14
Introduce option to customize generated docker files

### LINKED ISSUE #667: Setup/build-time instructions for Docker images
### Expected Behavior

I have a Docker image I'm building with the Micronaut Gradle Plugin. Specifically, it's an optimized native image, using the built-in GraalVM integration. It is generated in the following steps:

1) `optimizedDockerfileNative` is executed to build the dockerfile
2) `optimizedDockerPrepareContext` is executed to build the config dirs
3) `optimizedBuildNativeLayersTask` is executed to build the native layers

during this process, i copy a file, `server.iprof`, into the native image root, so that i can use [PGO](https://www.graalvm.org/22.0/reference-manual/native-image/PGO/) with my native image. the profile is generated ahead of time, so I just need to place it where Docker can find it, and execute a `COPY ...` instruction to make it available during subsequent native image builds.

so, i have added an `instruction` to do this:
```kotlin
optimizedDockerfileNative {
    graalImage.set("...")
    baseImage("...")
    args(*baseJvmArgs.plus(listOf(
      "...",
    )).toTypedArray())

    instruction("COPY server.iprof /home/app/server.iprof")
}
```

i expect this to be added at build-time, because all the runtime resources for the image are on the classpath, and embedded in the binary.

### Actual Behaviour

rather than injecting this instruction during the build stage of the Dockerfile, using this mechanism, the instruction is injected during the runtime assembly stage:
```dockerfile
FROM ... AS graalvm
WORKDIR /home/app
COPY layers/classes /home/app/classes
RUN mkdir /home/app/config-dirs
COPY config-dirs/generateResourcesConfigFile /home/app/config-dirs/generateResourcesConfigFile
COPY config-dirs/... /home/app/config-dirs/...

> RUN native-image ... --pgo=server.iprof < file needed here

FROM ...
EXPOSE 8080
COPY --from=graalvm /home/app/application /app/application

> COPY server.iprof /home/app/server.iprof < instruction injected here

ENTRYPOINT ["/app/application", "..."]
```

### Steps To Reproduce

1. try to inject an instruction at build time
2. observe that the [NativeImageDockerfile code](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java#L493) does not inject from user instructions at such time

### Environment Information

- Operating System: macOS Ventura 13.2.1
- JDK Version: 19 (GraalVM)

### Example Application

_No response_

### Version

3.8.7

melix:
This commit introduces the possibility to customize the dockerfiles which are generated before building the docker image. To do this, the user can either configure a single task, e.g:

```gradle
tasks.named("dockerfile") {
   ...
}
```

or all tasks at once:

```gradle
tasks.withType(DockerBuildOptions).configureEach {
   ...
}
```

Then a DSL is exposed which allows easily modifying the generated file, for example:

```gradle
editDockerfile {
    after('COPY layers/libs /home/app/libs') {
        insert('COPY server.iprof /home/app/server.iprof')
    }
}
```

The editor DSL allows narrowing down what part of the file to edit thanks to nested `before` and `after` calls.

Note that this is a "poor man's" solution to the problem of customizing generated files, because using the current `Dockerfile` task APIs it is extremely difficult or even impossible to alter the configuration after it's done.

The editor DSL is minimal for now, in the sense that it only supports strings, not patters, for simplicity.

Fixes #667


sgammon:
@melix thank you!

PR #675, opened 2023-03-17
Make tests run in parallel

melix:


melix:
Closing because it doesn't bring enough value for the complexity it introduces

PR #679, opened 2023-03-22
Copy the customized CRaC dockerfiles if they exist in the root

### LINKED ISSUE #678: Adding a custom Dockerfile for CRaC checkpointing doesn't work
We allow users to specify a dockerfile for creating the CRaC checkpoint.

It should be called `DockerfileCracCheckpoint` and reside in the root of the project.

This doesn't work, as the docker context location is set to build/docker/main, and then it cannot find the dockerfile (everything must be in the context directory)

We need to copy the file into the build/docker/main directory if it is detected

timyates:
Previously, we allowed users to create `Dockerfile` and/or `DockerfileCracCheckpoint` in the root of their project and we then would attempt to use these for the image generation.

This doesn't work.  The Docker plugin requires all Docker content to be in the same file tree.

This PR copies checks for the existence of these files, and copies them to the correct location instead of building the dockerfiles.

Fixes #678


melix (approved):

sdelamo:
@melix can we merge this and release a new patch of the gradle plugin 🙏🏻

PR #701, opened 2023-04-24
Add an option to not automatically add BOM

### LINKED ISSUE #681: Make it possible to allow users to define the Micronaut BOM (in order to achieve centralized version management)
### Feature description

Hi, 

Thanks for making Micronaut such a great framework with GraalVM support out of the box. 

From the Gradle Plugin implementation https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/cc84332f5635e3da7c71e81460659f41fd36ae2b/minimal-plugin/src/main/java/io/micronaut/gradle/PluginsHelper.java#L48-L60 

It looks like users are forced to either define `micronautVersion` in `gradle.properties` file or use the `micronaut {}` extension to declare the version, but I would like to declare the version in my own version catalogue in `gradle/libs.verisons.toml` or in my own `dependencies.constraints` block (which is possibly inside another custom [gradle platform](https://youtu.be/8044F5gc1dE?list=PLWQK2ZdV4Yl2k2OmC_gsjDpdIBTN0qqkE&t=208)) in order to achieve centralized version management.

So maybe one feasible solution would be using [`Configuration.defaultDependencies()`](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/Configuration.html#defaultDependencies-org.gradle.api.Action-)? (similar to the [Lombok Plugin implementation](https://github.com/freefair/gradle-plugins/blob/3709c51ac65a64fb94ae350eb08423ec8932e1eb/lombok-plugin/src/main/java/io/freefair/gradle/plugins/lombok/LombokBasePlugin.java#L23-L25)), because from the Gradle documentation, the default dependencies are only added when the configuration has no dependencies. This is perfect as the plugin do not need to worry about adding Micronaut BOM when users already added it. The only concern is if users add some other dependencies to  `MICRONAUT_BOMS_CONFIGURATION` configuration but not the Micronaut BOM. I think in this case, we could add a check and throw the exception.

melix:
This option has (at least) a couple use cases:

1. the issue described in #681, when a user wants to drive the Micronaut version from a catalog, or directly in their dependencies (although this is probably redundant with #694)

2. we use the Gradle plugins in our module builds, where it's going to look for the platform BOM, which is not necessarily what we want, since we want to use local versions of projects and possibly override transitive versions

Closes #681

### DIFF
--- a/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautComponentPlugin.java
+++ b/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautComponentPlugin.java
@@ -18,7 +18,6 @@ package io.micronaut.gradle;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
-import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.artifacts.dsl.DependencyHandler;
 import org.gradle.api.plugins.BasePlugin;
 import org.gradle.api.plugins.JavaPlugin;
@@ -37,7 +36,6 @@ import org.gradle.api.tasks.testing.Test;
 import java.io.File;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


melix:
Here's an example of how a Micronaut module build can be adapted to use this flag, on `micronaut-rabbitmq`:

```diff
diff --git a/buildSrc/src/main/groovy/io.micronaut.build.internal.rabbitmq-examples.gradle b/buildSrc/src/main/groovy/io.micronaut.build.internal.rabbitmq-examples.gradle
index d632e272..97e7fcfe 100644
--- a/buildSrc/src/main/groovy/io.micronaut.build.internal.rabbitmq-examples.gradle
+++ b/buildSrc/src/main/groovy/io.micronaut.build.internal.rabbitmq-examples.gradle
@@ -5,6 +5,9 @@ plugins {
 }
 
 dependencies {
+    micronautBoms(platform(mn.micronaut.core.bom))
+    micronautBoms(platform(mnTest.micronaut.test.bom))
+    micronautBoms(platform(mnTestResources.micronaut.test.resources.bom))
     testImplementation projects.micronautRabbitmq
     testImplementation mn.reactor
     testImplementation mnSerde.micronaut.serde.support
@@ -18,7 +21,8 @@ configurations.all {
     }
 }
 micronaut {
-    version libs.versions.micronaut.platform.get()
+    importMicronautPlatform = false
+    //version libs.versions.micronaut.platform.get()
     testRuntime "junit5"
     testResources {
         clientTimeout = 300
diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
index b421ff91..cd10d545 100644
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,8 +1,9 @@
 [versions]
 micronaut-docs = '2.0.0'
 micronaut = '4.0.0-M2'
-micronaut-platform = '4.0.0-M1'
+#micronaut-platform = '4.0.0-M1'
 micronaut-test = '4.0.0-M1'
+micronaut-test-resources = '2.0.0-M1'
 micronaut-gradle-plugin = "4.0.0-M1"
 groovy = '4.0.10'
 spock = '2.3-groovy-4.0'
@@ -36,6 +37,7 @@ testcontainers-spock = { module = 'org.testcontainers:spock', version.ref = 'tes
 
 micronaut-micrometer = { module = "io.micronaut.micrometer:micronaut-micrometer-bom", version.ref = "micronaut-micrometer" }
 micronaut-serde = { module = "io.micronaut.serde:micronaut-serde-bom", version.ref = "micronaut-serde" }
+micronaut-test-resources = { module = "io.micronaut.testresources:micronaut-test-resources-bom", version.ref = "micronaut-test-resources" }
 gradle-micronaut = { module = "io.micronaut.gradle:micronaut-gradle-plugin", version.ref = "micronaut-gradle-plugin" }
 gradle-kotlin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
 gradle-kotlin-allopen = { module = "org.jetbrains.kotlin:kotlin-allopen", version.ref = "kotlin" }
diff --git a/settings.gradle b/settings.gradle
index 4dbb2206..881afc18 100644
--- a/settings.gradle
+++ b/settings.gradle
@@ -26,4 +26,5 @@ micronautBuild {
     importMicronautCatalog()
     importMicronautCatalog("micronaut-micrometer")
     importMicronautCatalog("micronaut-serde")
+    importMicronautCatalog("micronaut-test-resources")
 }
```

Basically, the `micronaut.platform` solution has the advantage of being _easy to setup_, but you would still possibly add constraints which are not wanted when testing. By setting `importMicronautPlatform` to `false`, then the BOM is not applied anymore, so you have to explicitly import all the module BOMs that you are using, which provides correctness at the cost of simplicity.

PR #702, opened 2023-04-25
M2

sdelamo:


timyates:
I believe this is failing as it's looking for platform-M2, and we only have M1

PR #714, opened 2023-05-10
graalvm native-image docker replace ol7 with ol8

sdelamo:
Should we use `ol7` instead of `ol8` ? see:
https://github.com/micronaut-projects/micronaut-starter/pull/1807

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -373,7 +373,7 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
     }
 
     private static String toGraalVMBaseImageName(String graalVersion, String jdkVersion) {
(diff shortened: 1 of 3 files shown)


melix:
mmm, look at commit 23a44a49bd0abd3f7412f5a4e3517ed08a795050, we reverted to `ol7` precisely because of zip corruption...

timyates:
@alvarosanchez Should we also move to ol8 in the Maven plugin?

https://github.com/micronaut-projects/micronaut-maven-plugin/blob/master/micronaut-maven-plugin/src/main/java/io/micronaut/maven/AbstractDockerMojo.java#L161-L168

alvarosanchez:
> @alvarosanchez Should we also move to ol8 in the Maven plugin?

Absolutely. It really bothers me that we still don't have a central place to change this, and we always have to play catching up games to stay in sync

timyates:
Raised a PR to default to ol8 https://github.com/micronaut-projects/micronaut-maven-plugin/pull/723

PR #716, opened 2023-05-11
Remove snapshot repos

melix:
Now that we have milestones for everything, snapshots are no longer required. Kept as comments for the next major changes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,9 +24,9 @@ plugins {
 repositories {
     mavenCentral()
     gradlePluginPortal()
-    maven {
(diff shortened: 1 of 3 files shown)


alvarosanchez (approved):
I'm curious: can you not leverage the Groovy DSL and add a method that enables snapshot repos based on a property or something? I can't do that in XML 😅, but you can. That would be better than commenting out IMO

melix:
It is possible, the difficulty being that it's defined in several places. The issue with the DSL is _when_ the repo is added, though.

PR #728, opened 2023-05-24
Add core BOM

n0tl3ss:


melix:
Why are you adding this?

graemerocher:
it enables automatic renovate PRs

melix (changes requested):
The `micronaut` version here refers to the `platform` (see the comment on the version). It is only used in tests. The Gradle plugin doesn't depend on `micronaut-core`.

graemerocher:
perhaps it should be renamed as it is inconsistent with the rest of the modules

PR #729, opened 2023-05-25
Add Official CRaC JDK support to the Gradle Plugin.

timyates:
Azul has officially released their CRaC JDK.

This PR switches to find and download the official JDK when building with `dockerBuildCrac`.

They have also silently released the ARM version of the CRaC enabled JDK.

This PR also adds support for this, using the current system architecture as the default.

New configuration has been added to allow changing from the defaults.

The old "platform" configuration has been deprecated for removal in 4.0.0.

### DIFF
--- a/crac-plugin/src/main/java/io/micronaut/gradle/crac/CRaCConfiguration.java
+++ b/crac-plugin/src/main/java/io/micronaut/gradle/crac/CRaCConfiguration.java
@@ -3,6 +3,7 @@ package io.micronaut.gradle.crac;
 import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.PathSensitive;
 import org.gradle.api.tasks.PathSensitivity;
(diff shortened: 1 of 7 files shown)


sdelamo (approved):

sdelamo:
@melix can you review this?

crac-plugin/src/main/java/io/micronaut/gradle/crac/CRaCConfiguration.java
```diff
@@ -21,10 +21,26 @@ public interface CRaCConfiguration {
...
+     * Currently only {@value MicronautCRaCPlugin#X86_64_ARCH} and {@value MicronautCRaCPlugin#ARM_ARCH} are supported.
+     * @return the architecture
+     */
+    Property<String> getArch();
+
+    /**
+     * The java version to use for building the CRaC enabled images. Currently only '17' is supported.
+     * @return the java version
+     */
+    Property<String> getJavaVersion();
```
melix, line 35: I think we should use a `Property<JavaLanguageVersion>` here

melix (changes requested):

timyates:
@melix Switched to use JavaLanguageVersion instead of String 👍

timyates:
Did you mean JavaVersion? 🤔

melix (approved):

PR #736, opened 2023-06-05
Initial implementation of Micronaut OpenAPI plugin

melix:
This commit demonstrates an integration of the Micronaut OpenAPI generator directly into the Gradle plugin. Because that project is still work in progress, it uses a source dependency.

This feature allows the generation of an OpenAPI client or server directly in the build pipeline.

For now, we only support generation of apis and models. Tests are not supported, but it is arguable if we should handle this at the plugin level or via something like Micronaut launch instead.

When the `io.micronaut.openapi` plugin is applied, then a new `openapi` extension is added to the `micronaut` configuration block, letting the user declare a definition to use. For example:

```
micronaut {
   ...
   openapi {
       client(file("src/openapi/petstore.json")) {
           apiPackageName = "com.mycompany.petstore.api"
           modelPackageName = "com.mycompany.petstore.model"
       }
   }
}
```

Then the APIs and models are generated (in 2 distinct tasks) as part of the build process. We're using generated output directories (e.g `build/generated/openapi/client/...`) so it means that the tasks are executed on every invocation, but they are cacheable.

Relates to https://github.com/micronaut-projects/micronaut-openapi/issues/1037


melix:
@graemerocher this PR demonstrates the use of the new Micronaut OpenAPI generator in the Gradle plugin. I want to investigate a way to do something similar in the Maven plugin, since @alvarosanchez is quite busy already.

@andriy-dmytruk is updating the generator to use interfaces instead of abstract classes, and I will also need to make changes to expose more parameters (for example there's currently no way to select between the serde libraries in the generator API, although it supports both).

graemerocher (approved):

sdelamo (changes requested):
This looks great. Can you however clarify what does it mean to generate a server or a client in the documentation? 

> You can generate a server by configuring the 

Can you explain to the user what gets generated and where. Is it a class which the user should extend is it bytecode? is it a java class?

melix:
I have updated the docs

PR #748, opened 2023-06-16
Upgrade to GraalVM 23

melix:
Setting as draft PR because the Docker images are not yet available.

### DIFF
--- a/functional-tests/src/test/groovy/io/micronaut/gradle/lambda/LambdaNativeImageSpec.groovy
+++ b/functional-tests/src/test/groovy/io/micronaut/gradle/lambda/LambdaNativeImageSpec.groovy
@@ -93,7 +93,7 @@ class LambdaNativeImageSpec extends AbstractFunctionalTest {
         dockerfileNativeTask.outcome == TaskOutcome.SUCCESS
 
         and:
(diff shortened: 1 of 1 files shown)


breucode:
The reason, the images are not available is because graalvm changed their versioning.

The images you are looking for are located here:

https://github.com/graalvm/container/pkgs/container/native-image-community

sdelamo:
@melix can we move this forward?

sdelamo:
@breucode we will wait for the official Oracle images for GraalVM 23.

breucode:
> @breucode we will wait for the official Oracle images for GraalVM 23. 
> 
> 

I am a little confused. According to https://www.graalvm.org/faq/ the versioning scheme of graalvm was changed.

"As part of this alignment, GraalVM will adopt the JDK’s release numbering scheme based on the supported Java version. To avoid confusion with older releases, new releases will be named GraalVM for JDK <Java version>, for example GraalVM for JDK 20."

So the newest Releases would be GraalVM for JDK 17 and GraalVM for JDK 20.

This is also reflected here: https://www.graalvm.org/release-notes/

melix:
This is orthogonal to the available docker images. And you will still have GraalVM 23 for Java 17 for example.

melix:
Sorry I mixed up images with another project, using `ghcr.io/graalvm/native-image-community` will definitely work, I'm going to fix that PR accordingly.

PR #754, opened 2023-06-29
Add option to configure the native lambda runtime

### LINKED ISSUE #753: Ability to set mainClass for native Lambda
We need an option to provide the main class (the entry point for the native image) for AWS Lambda functions which use the _Default_ application type.  

We currently set it to  [io.micronaut.function.aws.runtime.MicronautLambdaRuntime](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/graalvm-plugin/src/main/java/io/micronaut/gradle/graalvm/MicronautGraalPlugin.java#L90)

However, we ship now serveral main classes depending on the event they need to support. 

A user may want to set: 

- [`io.micronaut.function.aws.runtime.MicronautLambdaRuntime`](https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-custom-runtime/src/main/java/io/micronaut/function/aws/runtime/MicronautLambdaRuntime.java)
- [`io.micronaut.function.aws.runtime.APIGatewayV2HTTPEventMicronautLambdaRuntime`](https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-custom-runtime/src/main/java/io/micronaut/function/aws/runtime/APIGatewayV2HTTPEventMicronautLambdaRuntime.java)
- [`io.micronaut.function.aws.runtime.ApplicationLoadBalancerMicronautLambdaRuntime`](https://github.com/micronaut-projects/micronaut-aws/blob/master/function-aws-custom-runtime/src/main/java/io/micronaut/function/aws/runtime/ApplicationLoadBalancerMicronautLambdaRuntime.java)

it would be nice we could do: 

```kotlin
micronaut {
    nativeLambda {
        mainClass.set("io.micronaut.function.aws.runtime.APIGatewayV2HTTPEventMicronautLambdaRuntime")
    }
}
```

melix:
Closes #753


minimal-plugin/src/main/java/io/micronaut/gradle/graalvm/NativeLambdaExtension.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Allows configuring the native lambda specific parameters.
+ */
+public interface NativeLambdaExtension {
+    /**
+     * The native lambda runtime
+     * @return the runtime
+     */
+    Property<NativeLambdaRuntime> getLambdaRuntime();
```
sdelamo, line 28: Would it be possible for this to be a String instead of an enum? That way it would be more extensible and flexible. Users may be able to write other main classes. 

We will generate in starter the appropriate  String.

sdelamo (changes requested):
Would it be

minimal-plugin/src/main/java/io/micronaut/gradle/graalvm/NativeLambdaExtension.java
melix: I think we can support both. I don't like stringly typed APIs much when we can do better. Especially here the user would have to figure out the main classes. Even if you help in starter, switching from one to the other would be more complicated.

What about having :

```
public interface NativeLambdaExtension {
    /**
     * The native lambda runtime
     * @return the runtime
     */
    Property<NativeLambdaRuntime> getLambdaRuntime();

    Property<String> getLambdaRuntimeClassName();
}
```

where the 2d property would let the user specify an explicit class name, but the default value would be the enum class?

melix: I pushed the changes I was thinking about, let me know

minimal-plugin/src/main/java/io/micronaut/gradle/graalvm/NativeLambdaRuntime.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.gradle.graalvm;
+
+public enum NativeLambdaRuntime {
+    STANDARD("io.micronaut.function.aws.runtime.MicronautLambdaRuntime"),
```
sdelamo, line 19: Please rename to `API_GATEWAY_V1`

minimal-plugin/src/main/java/io/micronaut/gradle/graalvm/NativeLambdaRuntime.java
```diff
@@ -0,0 +1,33 @@
...
+    API_GATEWAY("io.micronaut.function.aws.runtime.APIGatewayV2HTTPEventMicronautLambdaRuntime"),
```
sdelamo, line 20: Please rename to `API_GATEWAY_V2`

minimal-plugin/src/main/java/io/micronaut/gradle/graalvm/NativeLambdaRuntime.java
```diff
@@ -0,0 +1,33 @@
...
+    LOAD_BALANCER("io.micronaut.function.aws.runtime.ApplicationLoadBalancerMicronautLambdaRuntime");
```
sdelamo, line 21: Please rename to `ALB`

sdelamo (changes requested):

melix:
Constants renamed!

sdelamo (approved):
Thanks for doing this so quick, could you release a new version of the Micronaut Gradle plugin which includes this PR.

PR #805, opened 2023-08-18
Micronaut 4.0.1 `buildNativeLambda` task breaks under Windows with `StringIndexOutOfBoundsException`

guillermocalvo:
This PR fixes a bug introduced in #654 that affects only Windows.

The exclude-config arguments were being monkey-patched incorrectly: the last part of the regular expression `\Q ... \E` was being interpreted as a path separator on Windows. This PR makes `lastIndexOf` ignore that part.

No additional tests are needed, because `LambdaNativeImageSpec` was already failing on Windows due to this issue. I removed `@IgnoreIf({ os.windows })` and it works fine after the bugfix.


sdelamo (approved):

melix:
Fixed in #808

PR #807, opened 2023-08-20
OpenAPI generator new properties

altro3:
Prepare for next micronaut-openapi release


altro3:
@melix hi! Yeah, I'll finish tomorrow

altro3:
@melix oh, the latest release 5.1.1 didn't include these changes :-(. I'm afraid we'll have to wait for the 5.2.0 release...

melix:
Should be doable now that 5.2.0 is out.

altro3:
@melix Yes, that's right, but you need to add a language selection. Now there is a generator for Kotlin. I'll try to do it this week

melix:
This looks like a breaking change, right? There's now `forJavaServer` and `forKotlinServer`, instead of just `forServer`, so I guess older builds which use these will break. Was that change necessary?

altro3:
@melix I just tried not to change the old structure and integrate support for the new generator with minimal changes

altro3:
Generator settings for Kotlin and Java may differ. Yes, at the moment they coincide in this place, but in general, no one guarantees that this will not change later. The generators themselves for Java and Kotlin are absolutely not related to each other. They have different settings and different classes

altro3:
Yes, by the way, how can this be a breaking change if nothing changes for the user and the settings through the plugin are done in the same way as before?

melix:
Got it, but from what I see, openapi-generator has a breaking change. It would have been better to keep `forServer` as an alias for `forJavaServer`. Integrating such breaking changes will require bumping the major version of the plugin.

altro3:
What do you suggest? Just add `forServer` with `@Deprecated` annotation, which call `forJavaServer` method inside?

And why we can't up major version of plugin?

melix:
No you're right that it doesn't seem to break compatibility here since the mapping is done internally. I will checkout the branch locally and carefully review it, easier than GitHub. Note that we _can_ bump the plugin version but it's always a bit annoying because we maintain several branches.

melix:
@altro3 I have fixed the Kotlin tests setup. However it seems that the Kotlin generator generates incorrect Kotlin code: the server tests are failing with a compile error. It is also possible that we need to properly setup the dependencies for Kotlin sources, I'm not sure yet.

altro3:
@melix could you alsio update micronaut-openapi to version 6.2.0 ? @sdelamo released it right now. This version contains fixes for the latest known bugs. There's nothing else to fix for now

altro3:
@melix I checked the generator. The code is correct, apparently, the dependencies really need to be fixed. Then I'll look at it tomorrow and then I can update micronaut-openapi to the latest version myself

melix:
Ok, I'm confused because compilation fails on this line:
```kotlin
security = [
            SecurityRequirement(name = "petstore_auth", scopes = {"write:pets", "read:pets"})
        ]
```

with:

```
> Task :kaptGenerateStubsKotlin FAILED
e: file:///tmp/junit6466566742010923247/build/generated/openapi/generateServerOpenApiApis/src/main/kotlin/io/micronaut/openapi/api/PetApi.kt:51:79 Unexpected tokens (use ';' to separate expressions on the same line)
```

altro3:
> Ok, I'm confused because compilation fails on this line:
> 
> ```kotlin
> security = [
>             SecurityRequirement(name = "petstore_auth", scopes = {"write:pets", "read:pets"})
>         ]
> ```
> 
> with:
> 
> ```
> > Task :kaptGenerateStubsKotlin FAILED
> e: file:///tmp/junit6466566742010923247/build/generated/openapi/generateServerOpenApiApis/src/main/kotlin/io/micronaut/openapi/api/PetApi.kt:51:79 Unexpected tokens (use ';' to separate expressions on the same line)
> ```

I'll check or fix it tomorrow. Maybe it's really bug

altro3:
I think problem is here - `{"write:pets", "read:pets"}`, must be `["write:pets", "read:pets"]`

melix:
yeah that's why I'm confused, for me there's a bug in the generator

altro3:
Yeah, I agree. Will fix it tomorrow

altro3:
@melix could you check what is wrong now? I fixed generator bug

altro3:
@lucjross-favor Hi! thanks for help. Yes, I saw this message in logs, but the problem is that this message has nothing to do with the generator. Those. This is a potential bug in micronaut-openapi (in the swagger file generator by code), but this has nothing to do with the bug with micronaut-openapi-generator - the code generator by swagger file. So I don't think that's the problem

altro3:
@melix could you help?

melix:
I think the error is unrelated here. The test resources tests put high load on agents and sometimes it fails. Also 3rd party branches cannot publish build scans.

altro3:
@melix yes, yes

PR #828, opened 2023-09-13
Fix CRaC JDK download in docker container

timyates:
Azul added a JRE to their CRaC java deliverables.

The query wasn't specific enough to say we want a JDK, so we end up pulling the JRE if it comes back first in the response. The JRE doesn't include the JCmd command which we use for snapshotting...

This change adds `java_package_type=jdk` to the request URL so we just get the JDK.

### Done on the 3.7.x branch, to allow fixing Micronaut 3.x


melix (approved):

timyates:
The windows pipeline has some issue with graal 🤔 seems to be looking for a json file that's not there any more 🤔

melix:
The workflows on branch 3.7.x are out of sync, which explains the error on Windows. I guess we have to ignore it for now.

PR #830, opened 2023-09-14
Use an Arm compatible base image for Arm dockerBuildNative

### LINKED ISSUE #829: Micronaut dockerBuildNative does not build aarch64 artifact
### Expected Behavior

Hello,

I have tried basic example from Micronaut Launch with 0 added features.
My machine is MacBook Pro M1.

According to documentation:

The plugin will detect the host operating system architecture (based on the os.arch Java system property) and will install the corresponding GraalVM binary distribution inside the Docker image. This means that when running packaging from an X86_64 (Intel/AMD) machine, the produced native executable will be an amd64 binary, whilst on an ARM host (such as the new Mac M1) it will be an aarch64 binary.
https://micronaut-projects.github.io/micronaut-gradle-plugin/latest/#_architecture_considerations

### Actual Behaviour

I can still see that it builds amd64 artifact and I cannot run it on my machine:

docker image inspect --format "{{.ID}} {{.RepoTags}} {{.Architecture}}" demo
sha256:08b3dd3b4eca5a81666c326dc7e9edfe919171f031d04b98f0efc23fd77be3d9 [demo:latest] amd64

### Steps To Reproduce

1. Use https://micronaut.io/launch/
2. Version 4.1.1, Java, Gradle, Junit - no added features
3. Make sure your machine is a Mac with M1/M2 processor
4. Run dockerBuildNative
5. Inspect image via CLI or via Docker Desktop

### Environment Information

MacOS Ventura 13.5.2

openjdk 17.0.8 2023-07-18 LTS
OpenJDK Runtime Environment Corretto-17.0.8.7.1 (build 17.0.8+7-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.8.7.1 (build 17.0.8+7-LTS, mixed mode, sharing)

### Example Application

_No response_

### Version

4.1.1

timyates:
For both Arm and X86 we are using an alpine base image.

There is no Arm variant of Alpine, so this builds an ARM native binary, then puts it in an x86 base image.

This change switches to use `cgr.dev/chainguard/wolfi-base:latest` when the architecture is Arm.

This image was chosen due to it's size... There may be a better default choice out there...

Closes #829

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -724,7 +724,11 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
             if (strategy == DockerBuildStrategy.LAMBDA && baseImage == null) {
                 baseImage = "amazonlinux:2";
             } else if (baseImage == null) {
-                baseImage = "frolvlad/alpine-glibc:alpine-" + DefaultVersions.ALPINE;
+                baseImage = getGraalArch()
+                        .map(a -> a.equals(ARM_ARCH)
+                                ? "cgr.dev/chainguard/wolfi-base:latest"
+                                : "frolvlad/alpine-glibc:alpine-" + DefaultVersions.ALPINE)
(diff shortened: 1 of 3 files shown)


timyates:
CRaC tests are failing as we need the fix from https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/828

melix (approved):

Issue #803, opened 2023-08-15
Index out of bounds if main class is not in a package


axel22:
### Expected Behavior

No index out of bounds exception.

### Actual Behaviour

Index out of bounds exception.

### Steps To Reproduce

Write any Gradle project with the Micronaut plugin in which the main class is not in a package.

Example:

```
plugins {
    id("com.github.johnrengelman.shadow") version "8.1.1"
    id("io.micronaut.application") version "4.0.2"
    id("io.micronaut.aot") version "4.0.2"
}

version = "0.1"
group = "simple.niohttp.app"

repositories {
    mavenCentral()
}

dependencies {
    annotationProcessor("io.micronaut.serde:micronaut-serde-processor")
    implementation("io.micronaut.serde:micronaut-serde-jackson")
    implementation("io.micronaut:micronaut-http-server-netty")
    runtimeOnly("ch.qos.logback:logback-classic")
    testImplementation("io.micronaut:micronaut-http-client")
}

application {
    mainClass.set("NioHttpServer")
}

java {
    sourceCompatibility = JavaVersion.toVersion("17")
    targetCompatibility = JavaVersion.toVersion("17")
}

graalvmNative.toolchainDetection.set(false)
micronaut {
    runtime("netty")
    testRuntime("junit5")
    processing {
        incremental(true)
    }
    aot {
        // Please review carefully the optimizations enabled below
        // Check https://micronaut-projects.github.io/micronaut-aot/latest/guide/ for more details
        enableNativeImage.set(true)
        optimizeServiceLoading.set(true)
        convertYamlToJava.set(false)
        precomputeOperations.set(true)
        cacheEnvironment.set(true)
        optimizeClassLoading.set(true)
        deduceEnvironment.set(true)
        optimizeNetty.set(true)
    }
}
graalvmNative {
    binaries["optimized"].buildArgs.add("-H:+UseDeprecatedOldClassInitialization")
}
```

Error:

```
Caused by: java.lang.StringIndexOutOfBoundsException: Range [0, -1) out of bounds for length 13
        at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:55)
        at java.base/jdk.internal.util.Preconditions$1.apply(Preconditions.java:52)
        at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:213)
        at java.base/jdk.internal.util.Preconditions$4.apply(Preconditions.java:210)
[...]

PR #836, opened 2023-09-21
Fix index out of bounds if main class is not in a package

IgorDomagala:
This PR fixes error mentioned in **Issue #803** by handling `getTargetPackage()` during `configureAotDefaults` setting `"io.micronaut.aot.generated"` as target package when main class is in the default package.


melix (approved):

melix:
Thanks for the fix!

PR #843, opened 2023-09-25
Add the Micronaut GraalVM dependency consistently

melix:
In previous releases of the plugin, there was an inconsistency in how the `micronaut-graal` processor was added. In particular, the dependency was only added if the JVM _running the build_ was GraalVM, but only for Kotlin. For Java and Groovy, the dependency was added in any case.

In fact, the behavior for Kotlin was wrong, since there's no reason that the toolchain used to compile the app is the same as the one running the build. Therefore, this commit changes the behavior so that as soon as the `io.micronaut.graalvm` plugin is applied, then the dependency is added.

It is a slight behavioral change for users who do _not_ use GraalVM but who would still apply the `io.micronaut.application` plugin. Before this change, they wouldn't have a `micronaut-graalvm` annotation processor on classpath. After this change, they will, because the `application` plugin applies the `graalvm` plugin. It is however significantly easier to reason about and more correct.

### DIFF
--- a/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautKotlinSupport.java
+++ b/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautKotlinSupport.java
@@ -1,7 +1,6 @@
 package io.micronaut.gradle;
 
 import com.google.devtools.ksp.gradle.KspExtension;
-import io.micronaut.gradle.graalvm.GraalUtil;
 import io.micronaut.gradle.internal.AutomaticDependency;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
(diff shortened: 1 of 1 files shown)


timyates (approved):
`GraalUtil` seems to be a class in 2 locations

minimal-plugin: https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/374a4ba6f4a1efbe3e8fce73b8e36cb64161de83/minimal-plugin/src/main/java/io/micronaut/gradle/graalvm/GraalUtil.java
graalvm-plugin https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/374a4ba6f4a1efbe3e8fce73b8e36cb64161de83/graalvm-plugin/src/main/java/io/micronaut/gradle/graalvm/GraalUtil.java

I think only the minimal-plugin one is used here

https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/e2976397c4701ed06291d50d7622e634f5aa2c33/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java#L99-L102

I guess deleting the graal one would be a breaking change... 🤔

timyates:
And even then, we've seed Graal JDKs with missing properties so aren't detected by this...

Maybe we should just annotate the Graal plugin one as deprecated?

graemerocher (approved):

timyates:
I was talking about `GraalUtil`, doesn't matter 👍

Issue #863, opened 2023-11-07
Upgrade to Kotlin 1.9.20


melix:
### Issue description

In order to fully support Java 21

PR #864, opened 2023-11-08
Upgrade to Kotlin plugin 1.9.20

melix:
Close: #863


timyates (approved):
I think some of the sonar smells are valid (local variables unused, etc)

PR #873, opened 2023-11-14
Document that plugin adds annotation processors automatically,…

### LINKED ISSUE #870: Gradle plugin adds automatically annotation processors but this feature is not documented
https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/minimal-plugin/src/main/java/io/micronaut/gradle/PluginsHelper.java#L186-L203

https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/minimal-plugin/src/main/java/io/micronaut/gradle/PluginsHelper.java#L57-L61

wetted:
and automatic dependencies can be suppressed.

closes #870

### DIFF
--- a/src/docs/asciidoc/index.adoc
+++ b/src/docs/asciidoc/index.adoc
@@ -306,6 +306,17 @@ The https://plugins.gradle.org/plugin/io.micronaut.library[Micronaut library plu
 * Applies the https://search.maven.org/artifact/io.micronaut.platform/micronaut-platform[Micronaut Bill of Materials (BOM)]
 * Applies the `java-library` plugin
 * Configures annotation processing for the current language (Groovy, Java or Kotlin)
+* Adds annotation processors automatically for Micronaut modules
+** `io.micronaut.data:micronaut-data-processor`
+** `io.micronaut.jaxrs:micronaut-jaxrs-processor`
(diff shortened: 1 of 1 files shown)


src/docs/asciidoc/index.adoc
```diff
@@ -306,6 +306,17 @@ The https://plugins.gradle.org/plugin/io.micronaut.library[Micronaut library plu
 * Applies the https://search.maven.org/artifact/io.micronaut.platform/micronaut-platform[Micronaut Bill of Materials (BOM)]
 * Applies the `java-library` plugin
 * Configures annotation processing for the current language (Groovy, Java or Kotlin)
+* Adds annotation processors automatically for Micronaut modules
```
sdelamo, line 309: Can you create a section where the automatic addition of annotation processors is explained and link it from here with the cross reference asciidoc syntax `<<autAnnProcessors,Adds annotation processors automatically for Micronaut modules>>` . 

We need to say something ike when the plugin detects you have a dependency with the data group id , it adds the annotation processor. Check the code to see what it is doing.

sdelamo (changes requested):

sdelamo (approved):

PR #899, opened 2023-12-04
Add overloads for local files in OpenAPI extension

### LINKED ISSUE #887: Add ability to process multiple files for micronaut-openapi-generator
### Feature description

Now we can generate only single client or / and server like this:
```groovy
mironaut {
    openapi {
        client(file("...")) {
        }
        server(file("...")) {
        }
}
```
Need to do something like this:

Variant 1:
```groovy
mironaut {
    openapi {
       // this variant not so good, because we can't customize package
        client(file("..."), file("..."), file("..."), ...) {
        }
        server(file("..."), file("..."), file("..."), ...) {
        }
}
```

Variant 2:
```groovy
mironaut {
    openapi {
       // this variant not so good, because we need to duplicate all properties every time, but this is not critical
        clients {
            client(file("...")) {
            }
            client(file("...")) {
            }
            client(file("...")) {
            }
        }
        servers {
            server(file("...")) {
            }
            server(file("...")) {
            }
        }
}
```

Variant 3:
```groovy
mironaut {
    openapi {
        clients {
// common properties for clients
            common {
            }
            client(file("...")) {
            }
            client(file("...")) {
            }
            client(file("...")) {
            }
        }
        servers {
// common properties for servers
            common {
            }
            server(file("...")) {
            }
            server(file("...")) {
            }
        }
}
```

melix:
Fixes #887

### DIFF
--- a/openapi-plugin/src/main/java/io/micronaut/gradle/openapi/DefaultOpenApiExtension.java
+++ b/openapi-plugin/src/main/java/io/micronaut/gradle/openapi/DefaultOpenApiExtension.java
@@ -223,6 +223,16 @@ public abstract class DefaultOpenApiExtension implements OpenApiExtension {
         }
     }
 
+    @Override
+    public void server(String name, File definition, Action<? super OpenApiServerSpec> spec) {
+        server(name, project.getObjects().fileProperty().fileValue(definition), spec);
+    }
+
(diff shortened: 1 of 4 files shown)


timyates (approved):
😎 👍

PR #903, opened 2023-12-08
Create multi-layer docker files

### LINKED ISSUE #735: Have a separate layers for SNAPSHOT dependencies
### Feature description

As explained in [How are Jib applications layered?](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-are-jib-applications-layered), we could have a separated layer for the "SNAPSHOT" dependencies  from the other so to not update the "static" dependencies layer that does not change between 2 builds

I've been able to do it locally by overloading the **BuildLayersTask** task
```groovy
    withType<BuildLayersTask> {
        doLast {
            // moving SNAPSHOT into a different layer
            copy {
                from(outputDir.dir("libs"))
                    .include("*-SNAPSHOT.jar")
                    .into(outputDir.dir("snapshotsLibs"))
            }
            delete {
                delete(fileTree(outputDir.dir("libs")).matching {
                    include("*-SNAPSHOT.jar")
                })
            }
        }
    }

    withType<MicronautDockerfile> {
        copyFile("layers/snapshotsLibs", "/home/app/libs")
}
```

but I feel that would be a great feature to be included by default

melix:
This commit improves our docker file generation in order to use multiple layers for dependencies. Instead of having a single libs directory, we now have 3 layers which are copied to the `libs` directory:

- one for the transitive dependencies which are not snapshot
- one for the snapshot dependencies
- one for the project dependencies

In addition to the layer which contains the application jar. Moreover, the dockerfile creation now makes use of the `--link` option for COPY, which makes it even more efficient.

Fixes #735

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java
@@ -25,6 +25,8 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
+import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.file.RegularFile;
 import org.gradle.api.plugins.BasePlugin;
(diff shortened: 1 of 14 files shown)


alvarosanchez (approved):
I've created https://github.com/micronaut-projects/micronaut-maven-plugin/issues/926 to implement the same in the Maven plugin.

Speaking of which, one thing that the Maven plugin does is running the application in exploded format (copying the `.class` files and running the main class), instead of running the fat JAR. It's a tiny improvement, but every ms helps :)

alvarosanchez:
Copying is slower, but I'm talking about running the app.

melix:
Do you actually have evidence that running is faster? When I was working on Gradle we had benchmarks showing that the runtime performance wasn't better with files, especially on large files with many class files.

alvarosanchez:
I've always seen it here and there, this is the latest that I've read: https://spring.io/blog/2023/12/04/cds-with-spring-framework-6-1

melix:
This may be specific to Spring though, since it does a lot of classpath scanning which is expensive. In addition, CDS mandates jars...

docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java
```diff
@@ -63,19 +65,54 @@ public void apply(Project project) {
...
+
+    public static void createDependencyLayers(MicronautDockerImage image, Configuration configuration) {
+        var projectLibs = configuration.getIncoming()
+            .artifactView(view -> {
+                view.lenient(true);
+                view.componentFilter(ProjectComponentIdentifier.class::isInstance);
+            }).getFiles();
+        var snapshotLibs = configuration.getIncoming()
+            .artifactView(view -> {
+//                view.lenient(true);
```
sdelamo, line 84: is this commented out view.lenient(true) intentional?  if not set is it equivalent to `view.lenient(false)`?

sdelamo (approved):

PR #918, opened 2024-01-17
Fixed cache misses caused by inspectRuntimeClasspath tasks.

ribafish:
Fixed [cache misses](https://ge.solutions-team.gradle.com/c/4hozdl3nprs6i/ftrsnzocqiem6/task-inputs?cacheability=cacheable,overlapping-outputs,validation-failure&expanded=WyJoNXl2aWZ1NTN2c2lhLSQxIiwibWZ1Ymk1a2lua24zZy0kMSIsImpwd2drenJuZ3kzanctcnVudGltZWNsYXNzcGF0aCIsImdzY3NyM2ZkazJvcnUtcnVudGltZWNsYXNzcGF0aCJd&page=3#gscsr3fdk2oru) caused by inspectRuntimeClasspath tasks. 

The issue was that when the runtimeClasspath input was defined just as `@InputFiles`, the files had bytewise differences because of timestamps. When using `@Classpath` annotation, the files contents are used for cache fingerpriting - one caveat is that with `@Classpath`, the order of files also matters, that's why I also sorted the files before adding them to the runtimeClasspath.

These issues were affecting the cacheability of all users of these plugins, including the [micronaut-data](https://github.com/micronaut-projects/micronaut-data) project.


minimal-plugin/src/main/java/io/micronaut/gradle/ApplicationClasspathInspector.java
```diff
@@ -18,12 +18,7 @@
 import org.gradle.api.DefaultTask;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.file.RegularFileProperty;
-import org.gradle.api.tasks.CacheableTask;
-import org.gradle.api.tasks.InputFiles;
-import org.gradle.api.tasks.OutputFile;
-import org.gradle.api.tasks.PathSensitive;
-import org.gradle.api.tasks.PathSensitivity;
-import org.gradle.api.tasks.TaskAction;
+import org.gradle.api.tasks.*;
```
runningcode, line 21: Is this in accordance with the Micronaut project style?

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautComponentPlugin.java
```diff
@@ -300,7 +300,10 @@ private static TaskProvider<ApplicationClasspathInspector> registerInspectRuntim
             var javaPluginExtension = PluginsHelper.javaPluginExtensionOf(project);
             task.setGroup(BasePlugin.BUILD_GROUP);
             task.setDescription("Performs sanity checks of the runtime classpath to warn about misconfigured builds");
-            task.getRuntimeClasspath().from(project.getConfigurations().getByName(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME));
+            // Sort the runtimeClasspath before setting it since the order doesn't actually matter for inspection, however it does matter for caching.
+            task.getRuntimeClasspath().from(project.getConfigurations()
+                    .getByName(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME).resolve()
```
runningcode, line 305: does this cause eager dependency configuration? that could be bad for performance.

minimal-plugin/src/main/java/io/micronaut/gradle/ApplicationClasspathInspector.java
melix: No, this should be restored as individual imports

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautComponentPlugin.java
melix: :x: This has indeed to be changed, configurations should never be resolved eagerly. For this check it doesn't really matter anyway to have sorted entries.

melix (changes requested):

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautComponentPlugin.java
ribafish: That is correct, however order matters for calculating the task inputs fingerprint, which is used for caching. I'll fix this ofc.

melix: If the order is different then it means that the graph is different, so to me it makes sense that the task is out of date.

runningcode (approved):

melix:
> n order to get the names of files on classpath, that classpath and its dependencies need to be resolved. With the current implementation that happens in execution time.

Yes, but we should change that to happen at configuration time, using a new `@Input` which uses the configuration resolved component api (which should be compatible with config cache too).

ribafish:
> > n order to get the names of files on classpath, that classpath and its dependencies need to be resolved. With the current implementation that happens in execution time.
> 
> Yes, but we should change that to happen at configuration time, using a new `@Input` which uses the configuration resolved component api (which should be compatible with config cache too).

Why would we want to have dependency resolution happen in configuration time? What are we trying to solve with this suggestion?

melix:
> Why would we want to have dependency resolution happen in configuration time? What are we trying to solve with this suggestion?

Correctness. With runtime classpath normalization, the task will be out of date if any if the files change. We don't care that they change, we only need the file names.

melix:
> By forcing configuration time dependency resolution you will do that on every build of every project that uses the plugin, even if the task is not executed

That is not correct. It would be done only if the task is on the task graph, thanks, hopefully, to lazy task configuration. If a task is on the task graph, then it needs to be executed hence configured. If the task is configured and not executed, then it's an error which needs fixing.

ribafish:
> > By forcing configuration time dependency resolution you will do that on every build of every project that uses the plugin, even if the task is not executed
> 
> That is not correct. It would be done only if the task is on the task graph, thanks, hopefully, to lazy task configuration. If a task is on the task graph, then it needs to be executed hence configured. If the task is configured and not executed, then it's an error which needs fixing.

I worded that badly - dependency resolution in configuration time will run every time the task is on the task graph, even if it doesn't need to be executed due to UP-TO-DATE checks or being retrieved FROM-CACHE. The task itself is attached to all `Test` tasks, so anytime any test is executed, the task will be on the task graph. Most of the time, the classpath wouldn't be changed, but the classpath would still be resolved in config time, even if the task execution could be avoided.

melix (approved):

ribafish:
@melix, I updated the PR as requested.

On a side-note, I was thinking that perhaps the best solution for this task would be to make it not cacheable. The execution is super fast since it just goes over the filenames.
- [Not cacheable](https://ge.solutions-team.gradle.com/s/wrofxtjl7r37e/timeline?hide-timeline&type=io.micronaut.gradle.ApplicationClasspathInspector&view=by-type)  - total time 3.865s
- [From cache with internal classpath ](https://ge.solutions-team.gradle.com/s/ecomcdf5kxdbw/timeline?hide-timeline&type=io.micronaut.gradle.ApplicationClasspathInspector&view=by-type)(PR above) - total time 5.2s, -0.04s avoidance savings with local cache
- [From cache with @Classpath](https://ge.solutions-team.gradle.com/s/f2ffo3sdbrekw/timeline?hide-timeline&type=io.micronaut.gradle.ApplicationClasspathInspector&view=by-type)  - total time 3.931s, -0.21s avoidance savings with local cache.

For reference, [this](https://ge.micronaut.io/s/sggaiuorf4orq/timeline?details=cnkx6czggr5qa&expanded=WyI3Il0&hide-timeline&type=io.micronaut.gradle.ApplicationClasspathInspector&view=by-type) is a real execution on [ge.micronaut.io](http://ge.micronaut.io/) - a lot of the time is spent dealing with remote cache and so all the tasks together take 14s.

melix:
Interesting numbers, indeed since most of the time will be spent locally dealing with dependency resolution, it's not worth caching.

PR #927, opened 2024-01-30
Use Chainguard's Wolfi base image by default

alvarosanchez:
This PR uses `cgr.dev/chainguard/wolfi-base:latest` as the default base image for all CPU architectures, not only for ARM, since it is multi-platform and a bit smaller than `frolvlad/alpine-glibc`

### DIFF
--- a/samples/test-resources/data-mysql/src/main/resources/application.yml
+++ b/samples/test-resources/data-mysql/src/main/resources/application.yml
@@ -16,3 +16,7 @@ datasources:
 #    driverClassName: org.postgresql.Driver
 #    schema-generate: CREATE_DROP
 #    dialect: POSTGRES
+test-resources:
+  containers:
+    mysql:
+      image-name: mysql:8.2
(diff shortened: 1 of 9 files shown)


melix (approved):

samples/test-resources/data-mysql/src/main/resources/application.yml
```diff
@@ -16,3 +16,7 @@ datasources:
 #    driverClassName: org.postgresql.Driver
 #    schema-generate: CREATE_DROP
 #    dialect: POSTGRES
+test-resources:
```
radovanradic, line 19: Sorry for interfering, just to ask can you try using test-containers 1.19.4 in this PR, maybe these mysql overrides won't be needed?
https://github.com/testcontainers/testcontainers-java/releases/tag/1.19.4 contains fix for MySQL it seems.

melix: Yes, I plan to test that version, but not in that PR, reason being that it's driven by test resources, not the Gradle plugin.

radovanradic: Ah, sorry, I assume it is test-resources as soon as I see your name:)

Issue #874, opened 2023-11-17
Support amazonlinux:2023 for built as a GraalVM Native executable to AWS Lambda Custom Runtime


liefra:
### Feature description

Would be great if the new [Amazon Linux 2023 runtime for AWS Lambda](https://aws.amazon.com/blogs/compute/introducing-the-amazon-linux-2023-runtime-for-aws-lambda/) is also supported.

## Current issue

When I change the config to:
```
tasks.named<io.micronaut.gradle.docker.NativeImageDockerfile>("dockerfileNative") {
    baseImage.set("amazonlinux:2023")
    args(
        "-XX:MaximumHeapSizePercent=80",
        "-Dio.netty.allocator.numDirectArenas=0",
        "-Dio.netty.noPreferDirect=true"
    )
}
```

Then a `gradlew dockerBuildNative` leads to:

```
> Task :dockerBuildNative
Building image using context '/Users/liefra/Downloads/demo/build/docker/native-main'.
Using Dockerfile '/Users/liefra/Downloads/demo/build/docker/native-main/DockerfileNative'
Using images 'demo'.
Step 1/38 : FROM amazonlinux:2023 AS graalvm
 ---> 23204fdc6e41
Step 2/38 : ENV LANG=en_US.UTF-8
 ---> Using cache
 ---> 05c10a24c84b
Step 3/38 : RUN yum install -y gcc gcc-c++ libc6-dev zlib1g-dev curl bash zlib zlib-devel zlib-static zip tar gzip
 ---> Running in fc4876439787
Failed to set locale, defaulting to C.UTF-8
Amazon Linux 2023 repository                    749 kB/s |  20 MB     00:27    
Last metadata expiration check: 0:00:14 ago on Fri Nov 17 09:06:23 2023.
No match for argument: libc6-dev
No match for argument: zlib1g-dev
Package bash-5.2.15-1.amzn2023.0.2.aarch64 is already installed.
Package zlib-1.2.11-33.amzn2023.0.4.aarch64 is already installed.
Error: Unable to find a match: libc6-dev zlib1g-dev


> Task :dockerBuildNative FAILED
11 actionable tasks: 3 executed, 8 up-to-date

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dockerBuildNative'.
[...]

PR #929, opened 2024-02-01
Update package names to support Amazon Linux 2023

wyattconsulting:
Unfortunately, libc6-dev and zlib1g-dev are not available in the Amazon Linux 2023 repository. Instead install the zlib-devel and glibc-devel packages, which are equivalent to zlib1g-dev and libc6-dev, respectively. This should be backwards compatible with Amazon Linux 2 and will also allow for using the amazonlinux:latest tag.

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
@@ -430,7 +430,7 @@ public abstract class NativeImageDockerfile extends Dockerfile implements Docker
         if (buildStrategy == DockerBuildStrategy.LAMBDA) {
             from(new From(imageResolver.resolve()).withStage("graalvm"));
             environmentVariable("LANG", "en_US.UTF-8");
-            runCommand("yum install -y gcc gcc-c++ libc6-dev zlib1g-dev curl bash zlib zlib-devel zlib-static zip tar gzip");
+            runCommand("yum install -y gcc gcc-c++ curl bash zlib zlib-devel zlib-static zip tar gzip");
             String jdkVersion = getJdkVersion().get();
             String graalArch = getGraalArch().get();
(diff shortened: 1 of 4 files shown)


wyattconsulting:
Sorry, not a regular contributor and so not sure how to indicate this is a proposed resolution for https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/874.

sdelamo (commented):
I am trying to test this change with both with AmazonLinux 2 and Amazon Linux 2023. if it works, we would need to change  in this PR the occurrences of `amazonlinux:2` with `amazonlinux:2023` to default to 2023.

wyattconsulting:
Thank you @sdelamo  for the testing. I started to setup a proper environment and realized it has been a very long time since I have done that. I updated the other references in the project and double-checked amazonlinux:2023 is the correct tag.

melix:
Looks like tests are failing with:

> Caused by: com.github.dockerjava.api.exception.DockerClientException: Could not build image: The command '/bin/sh -c yum install -y gcc gcc-c++ glibc-devel curl bash zlib zlib-devel zlib-static zip tar gzip'

wyattconsulting:
Looks like some sort of collision between curl and the curl-minimal on the image. This works:

`docker run --rm --entrypoint /bin/sh amazonlinux:2.0.20240109.0 -c "yum install -y gcc gcc-c++ glibc-devel curl bash zlib zlib-devel zlib-static zip tar gzip"`

and this works:

`docker run --rm --entrypoint /bin/sh amazonlinux:2.0.20240109.0 -c "yum install -y gcc gcc-c++ glibc-devel bash zlib zlib-devel zlib-static zip tar gzip"`

Will submit an update, but if amazonlinux:2 breaks without curl then looks like a breaking change might be required.

wyattconsulting:
`curl-minimal` seems to be the lowest common denominator

wyattconsulting:
Thank you, @melix

sdelamo (commented):
I have verified this works with Amazon Linux 2023 and with Amazon Linux 2.

sdelamo (approved):

PR #939, opened 2024-02-05
Restore "resources" layer

### LINKED ISSUE #938: Version 4.3.0 removed the resources layer
### Expected Behavior

Hey :wave: 

first of all thank you for your work. Sadly we have an issue with the recent updated to version 4.3.0.

Until version 4.2.1 there was a separate `resources` layer in the Dockerfile. This layer is still mentioned in the [latest guide](https://micronaut-projects.github.io/micronaut-gradle-plugin/latest/#_layered_application):

> The plugin creates a "layered" application in build/docker/main/layers and from that directory you can run java -jar myapp.jar. It works because that directory contains a lib directory with all the libraries and a resources directory with the configuration. Keep in mind that copying the only .jar file to another directory won’t work.

I would expect the `resource` layer to be still present or the guide to get updated with the recommended steps to create a container for an application that requires resource files.

### Actual Behaviour

With the update to 4.3.0 this layer got removed by PR #903. With this update our application isn't able to run from the created Docker container, because the container doesn't contain the `application.yml`, `logback.xml` or Flyway migration scripts.

I couldn't find a migration guide or any hint how this case is supported with the new version.

### Steps To Reproduce

_No response_

### Environment Information

_No response_

### Example Application

_No response_

### Version

4.3.0

melix:
This layer is actually required for regular docker images (no optimized/AOT) since the jars which are included are not the regular jars.

Fixes #938

### DIFF
--- a/docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java
+++ b/docker-plugin/src/main/java/io/micronaut/gradle/docker/MicronautDockerPlugin.java
@@ -71,6 +71,11 @@ public class MicronautDockerPlugin implements Plugin<Project> {
                 layer.getLayerKind().set(LayerKind.APP);
                 layer.getFiles().from(runnerJar);
             });
+            image.addLayer(layer -> {
+                layer.getLayerKind().set(LayerKind.EXPANDED_RESOURCES);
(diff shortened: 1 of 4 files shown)


sdelamo (approved):

sdelamo:
@melix remember to use GitHub Projects. I added it to 4.3.0.

PR #948, opened 2024-02-29
bug: copy layers into final crac image

timyates:
For some reason, we are no longer copying the libs and resources into the final CRaC image.

This PR remedies that


sdelamo (approved):

timyates:
I tried to add a functional test, but it couldn't find the crac plugin 🤔

melix (approved):

PR #508, opened 2022-07-20
Add support for AppCDS

melix:
Support application class data sharing by upgrading to Micronaut Test Resources 1.0.1.
This PR also renames the `testresources` configuration to `testResourcesService` for clarity (in a backwards compatible way). This should have been done in 3.5.0 but for some reason slipped out of the release.

PR #1004, opened 2024-06-05
Rename testresources to testResourcesService in doc

MT-Jacobs:
Per https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/508:

> This PR also renames the `testresources` configuration to `testResourcesService` for clarity (in a backwards compatible way). T


melix:
Good catch, thank you!

PR #1033, opened 2024-09-13
Update micronaut core version

andriy-dmytruk:
All the failing tests fail with:

```
Condition not satisfied:

result.output.contains('Creating bean classes for 1 type elements')
```
It seems directly related to this change: https://github.com/micronaut-projects/micronaut-core/pull/10769, which removes these exact logs. I will try fixing that, as I need the version for POJA changes.


melix (approved):
The removal of the "Creating bean classes" assertions is a bit annoying, but we can assume that it's well tested now...

PR #1039, opened 2024-10-14
feat: support using community graal for building native images

gphillips3-chwy:
The GraalVM version for building native images cannot be pulled directly from the community releases repository even when overriding the default `graalReleasesUrl` parameter.  This change introduces a new optional parameter `useGraalCommunityEdition` that when set to true will point to the [GraalVM community releases URL](https://github.com/graalvm/graalvm-ce-builds/releases) and construct a valid download URL for native image builds.


melix:
Given that builds started to fail because GraalVM wasn't available anymore at the URL we expected, next release will simply include the ability to specify the full URL to GraalVM. This way you can override with whatever version you need.

Thanks for the PR though!

PR #1041, opened 2024-10-22
Fix graalvm download url

melix:


alvarosanchez (approved):

docker-plugin/src/main/java/io/micronaut/gradle/docker/NativeImageDockerfile.java
```diff
@@ -431,13 +460,13 @@ private void setupInstructions(List<Instruction> additionalInstructions) {
         if (buildStrategy == DockerBuildStrategy.LAMBDA) {
             from(new From(imageResolver.resolve()).withStage("graalvm"));
             environmentVariable("LANG", "en_US.UTF-8");
-            runCommand("yum install -y gcc gcc-c++ glibc-devel glibc-langpack-en curl-minimal bash zlib zlib-devel zlib-static zip tar gzip");
+            runCommand("dnf update -y && dnf install -y gcc glibc-devel zlib-devel libstdc++-static tar && dnf clean all && rm -rf /var/cache/dnf");
```
nathanthorpe, line 463: I'm getting an error message here when building now, looks like it wants `gzip`

```
Step 5/40 : RUN tar -zxf /tmp/graalvm-jdk-21_linux-aarch64_bin.tar.gz -C /tmp && ls -d /tmp/graalvm-jdk-21* | grep -v "tar.gz" | xargs -I_ mv _ /usr/lib/graalvm
 ---> Running in b86271e6df5a
tar (child): gzip: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
```

nathanthorpe: Nevermind, it looks like the image was changed to `2023-minimal`, I had it set to `amazonlinux:2023`. Works with the new image.

alvarosanchez: @melix since the Gradle plugin allows changing this base image, [you should change it to install `gzip`](https://github.com/micronaut-projects/micronaut-starter/issues/2646#issuecomment-2470578477). That will work for both images.

Note that the change to the minimal image is intentional and more correct, since it is what [AWS uses in the provided runtime](https://docs.aws.amazon.com/linux/al2023/ug/lambda.html).

PR #1064, opened 2025-01-03
Gradle Configuration Cache fails to caches run task

### LINKED ISSUE #1063: Gradle Configuration Cache fails to caches run task
### Expected Behavior

To allow for better build performance, the run task should be compatible with the new [Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html).

### Actual Behaviour

When running a micronaut application with `gradle run` using the `io.micronaut.application:4.4.4` plugin, gradle reports an incompatibility at the end of the build (after stopping the application):

```
[error] cannot serialize object of type `org.gradle.api.internal.tasks.DefaultSourceSet`, a subtype of `org.gradle.api.tasks.SourceSet`, as these are not supported with the configuration cache.
    - field `val$sourceSet` of `io.micronaut.gradle.MicronautMinimalApplicationPlugin$1`
        - bean of type `io.micronaut.gradle.MicronautMinimalApplicationPlugin$1`
            - field `action` of `org.gradle.api.internal.AbstractTask$TaskActionWrapper`
                - bean of type `org.gradle.api.internal.AbstractTask$TaskActionWrapper`
                    - field `actions` of `org.gradle.api.tasks.JavaExec`
                        - task `:backend:run` of type `org.gradle.api.tasks.JavaExec`
```

### Steps To Reproduce

1. Create a Micronaut Application using the `io.micronaut.application` plugin
2. Enable the configuration cache (add `org.gradle.configuration-cache=true` to `gradle.properties`)
3. Run the application (`gradle run -t`)
4. Stop the application

I will follow up with a PR.

### Environment Information

- OS: Windows 10 64bit
- JDK: Azul 22.0.2 64bit

### Example Application

_No response_

### Version

4.7.1

Argannor:
Fixes #1063 

As the [gradle documentation](https://docs.gradle.org/8.10/userguide/configuration_cache.html#config_cache:requirements:disallowed_types) states:

> Gradle model types (e.g. Gradle, Settings, Project, SourceSet, Configuration etc…​) are usually used to carry some task input that should be explicitly and precisely declared instead.

> For example, if you reference a Project in order to get the project.version at execution time, you should instead directly declare the project version as an input to your task using a Property<String>. Another example would be to reference a SourceSet to later get the source files, the compilation classpath or the outputs of the source set. You should instead declare these as a FileCollection input and reference just that.

I verified the changes by publishing the new version to my local maven registry and following the reproduction steps outlined in the issue. Unfortunately I'm unable to verify the changes to the google cloud function runtime.

Please let me know if any further work is required!

### DIFF
--- a/functional-tests/src/test/groovy/io/micronaut/gradle/RuntimeDependenciesSpec.groovy
+++ b/functional-tests/src/test/groovy/io/micronaut/gradle/RuntimeDependenciesSpec.groovy
@@ -23,7 +23,7 @@ class RuntimeDependenciesSpec extends AbstractEagerConfiguringFunctionalTest {
             }
             
             micronaut {
-                version "$shadowVersion"
+                version "$micronautVersion"
                 runtime "$runtime"                
             }
             
--- a/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
+++ b/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
@@ -114,13 +114,13 @@ public class MicronautMinimalApplicationPlugin implements Plugin<Project> {
                     Map<String, Object> sysProps = new LinkedHashMap<>();
                     sysProps.put("micronaut.io.watch.restart", true);
                     sysProps.put("micronaut.io.watch.enabled", true);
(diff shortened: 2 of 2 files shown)


melix (approved):
Thanks, I think I will need to rework the Cloud Function anyway, the explicit dependsOn is a sign that something is not properly wired. Thanks for the fix!

PR #1081, opened 2025-02-19
Migrate to new shadow plugin `com.gradleup.shadow` (https://github.com/GradleUp/shadow)

### LINKED ISSUE #1025: Add support for the new Shadow Plugin/Group
### Feature description

Shadow plugin now has a new Maven Group and has released a new version supporting J21.

[Source](https://github.com/GradleUp/shadow/releases)

If we change the plugin name, it doesn't copy over all of the existing `META-INF/services` from the Micronaut Framework

Source: https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/c38e02635c51bd8d4df928a8d324e205e0676ec0/minimal-plugin/src/main/java/io/micronaut/gradle/ShadowPluginSupport.java#L37

### LINKED ISSUE #1080: Migrate to new Shadow Gradle Plugin iD
### Feature description

Per note in Gradle Shadow

https://github.com/GradleUp/shadow?tab=readme-ov-file#gradle-shadow
> [!NOTE]\
> Previously this plugin was developed by [@johnrengelman](https://github.com/johnrengelman) and published under the ID [`com.github.johnrengelman.shadow`](https://plugins.gradle.org/plugin/com.github.johnrengelman.shadow)
> before maintenance was transferred to the [GradleUp organization](https://github.com/GradleUp) to ensure future development, see [#908](https://github.com/GradleUp/shadow/issues/908).
>
> If you are still using the old plugin ID in your build script, we recommend to switch to the new plugin ID [`com.gradleup.shadow`](https://plugins.gradle.org/plugin/com.gradleup.shadow)
> and update to the latest version to receive all the latest bug fixes and improvements.

altro3:
Fixed #1080
Fixed #1025

### DIFF
--- a/functional-tests/src/test/groovy/io/micronaut/gradle/FullMicronautRuntimeSpec.groovy
+++ b/functional-tests/src/test/groovy/io/micronaut/gradle/FullMicronautRuntimeSpec.groovy
@@ -13,7 +13,7 @@ class FullMicronautRuntimeSpec extends AbstractEagerConfiguringFunctionalTest {
         buildFile << """
             plugins {
                 id "io.micronaut.application"
-                id("com.github.johnrengelman.shadow") version("$shadowVersion")
+                id("com.gradleup.shadow") version("$shadowVersion")
             }
             
             micronaut {
--- a/functional-tests/src/test/groovy/io/micronaut/gradle/RuntimeDependenciesSpec.groovy
+++ b/functional-tests/src/test/groovy/io/micronaut/gradle/RuntimeDependenciesSpec.groovy
@@ -19,7 +19,7 @@ class RuntimeDependenciesSpec extends AbstractEagerConfiguringFunctionalTest {
         buildFile << """
             plugins {
                 id "io.micronaut.minimal.application"
-                id("com.github.johnrengelman.shadow") version("$shadowVersion")
+                id("com.gradleup.shadow") version("$shadowVersion")
             }
             
             micronaut {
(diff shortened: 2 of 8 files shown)


melix (changes requested):
Thanks for the PR. I think we should make sure we support both the old plugin and the new one. There are simply too many builds using the old version for us to ignore, so upgrading the Micronaut Gradle plugin would be a potential breaking change, so we should add tests showing it works with both.

altro3:
I don't agree, becuse this is not a different plugin. All what was changd - gradle plugin ID and maven group. All packages ? classes and code are the same. This is just a next version of shadow plugin. All what need to change for develeopers, who wil use next release of micronaut gradle plugin - just change `id("com.github.johnrengelman.shadow") version '8.1.1'` to `id("com.gradleup.shadow") version '8.3.6'` and that's all changes

melix:
You don't agree but I think we should. Nothing prevents the gradlex version from breaking compatibility in the future, for example. It doesn't take much to add a test.

altro3:
Ok, but what we need to test? I mean I don't understand how test must be should look

altro3:
@melix done. Now we have 2 tests for both plugins

melix:
Great, thank you sir!

melix:
@altro3 Actually just found that this breaks builds with AOT. The change in `ShadowPluginSupport` makes it so that we don't see when the old plugin is applied. We should handle both ids.

PR #1124, opened 2025-06-07
Micronaut OpenAPI 6.19.0

altro3:


openapi-plugin/src/main/java/io/micronaut/gradle/openapi/MicronautOpenApiPlugin.java
```diff
@@ -23,7 +23,7 @@
 
 import java.util.List;
 
-public class MicronautOpenAPIPlugin implements Plugin<Project> {
+public class MicronautOpenApiPlugin implements Plugin<Project> {
```
melix, line 26: :x: This is a breaking change. Please revert this, as it is a public class and nothing prevents another plugin from applying it by class literal. This is in fact the recommended approach when composing plugins.

melix (changes requested):
Thanks, it's a rather involved PR! I noticed one breaking change, but in terms of behavior, it might also be the case. I didn't spot any convention change, but can you confirm that the defaults haven't changed?

openapi-plugin/src/main/java/io/micronaut/gradle/openapi/MicronautOpenApiPlugin.java
altro3: @melix ok, got it! Reverted

altro3:
About convention: In general, at the moment I have supplemented the convention so that all values ​​are present there. That is, I have not changed it, but expanded it. At the same time, just in case, I have replaced the get method with getOrNull , because if Null is passed, the generator will take the default value that is hardcoded into its code and there will be no error if the value is not specified or we accidentally missed it in the convention

altro3:
@melix The same problem as here: https://github.com/micronaut-projects/micronaut-openapi/pull/2183

We can no longer rely on the gradle to select the test launcher. What we need to change now:

1. There is no longer any point in using `useJUnitPlatform()`, because we need the launcher and engine versions to match

2. We now need to explicitly specify the dependency
```kotlin
testRuntimeOnly(mnTest.junit.platform.launcher)
```
I think it would be cool to add this dependency inside the `io.micronaut.build.shared.settings` plugin (for micronaut projects) and in these plugins for users: `io.micronaut.minimal.application`, `io.micronaut.application`, `io.micronaut.library`

altro3:
Links with information about this:


https://docs.gradle.org/8.12/userguide/upgrading_version_8.html#test_framework_implementation_dependencies

https://youtrack.jetbrains.com/issue/IDEA-370750/Idea-is-incompatible-with-JUnit-5.12

https://dev.to/be-hase/important-notes-on-junit-5120-in-gradle-13fj

Issue #1109, opened 2025-04-02
Add built-in java http server as a runtime supported by the gradle plugin


sdelamo:
### Feature description

see: https://micronaut-projects.github.io/micronaut-servlet/5.2.1/guide/#httpServer

PR #1125, opened 2025-06-10
Built-In Java HTTP Server Runtime

sdelamo:
https://micronaut-projects.github.io/micronaut-servlet/latest/guide/#httpServer

Close: https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/1109


sdelamo:
@melix I don't think the failures are related to my PR.

melix:
Indeed they seem unrelated. Not sure why these tests suddenly fail :thinking:

melix:
I suspect the failures are accidentally highlighting a real bug in the json schema generator, and are caused by the Micronaut dependency update. Can you do the upgrade in a separate PR, so that we can merge this one?

PR #1130, opened 2025-06-17
Configure prepareJitOptimizations task to use configured Java toolchain, fixes #1005

### LINKED ISSUE #1005: AOT is not aware of the active Java toolchain
### Expected Behavior

Running `prepareNativeOptimizations` or `prepareJitOptimizations` should use the same Java toolchain assigned to `javaCompile`, to avoid issues related to pinned JDKs

### Actual Behaviour

`AbstractMicronautAotCliTask` uses `getExecOperations().javaexec(...)` without awareness of toolchains; there is no way to assign the `javaLauncher` or `javaCompiler` for a `AbstractMicronautAotCliTask`, and so the user has a choice: disable toolchains and rely on `JAVA_HOME`, or live with the toolchain the task selects.

In my case, I need GraalVM so that I can pre-optimize my app which uses Truffle. Truffle needs a GraalVM JDK, or a `javac`/`java` run enabled with `-XX:+EnableJVMCI` and `--upgrade-module-path`. I tried this approach, but was unable to resolve the JAR needed for the GraalVM compiler since Gradle runs the AOT task configuration quite early in the build graph.

Ultimately, the user should have the ability to control the toolchain for the AOT tasks.

### Steps To Reproduce

See above

### Environment Information

- Operating System: macOS Sonoma 14.4.1
- JVM: See below

```
java version "22.0.1" 2024-04-16
Java(TM) SE Runtime Environment Oracle GraalVM 22.0.1+8.1 (build 22.0.1+8-jvmci-b01)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 22.0.1+8.1 (build 22.0.1+8-jvmci-b01, mixed mode, sharing)
```

### Example Application

_No response_

### Version

4.4.0

araysse:


melix:
This includes changes from your previous PR. Can you rebase onto `master`? Thanks!

aot-plugin/src/main/java/io/micronaut/gradle/aot/AbstractMicronautAotCliTask.java
```diff
@@ -42,6 +43,7 @@
 
 abstract class AbstractMicronautAotCliTask extends DefaultTask implements OptimizerIO {
 
+    private final Property<JavaLauncher> javaLauncher;
```
melix, line 46: :x: Use `public abstract Property<JavaLauncher> getJavaLauncher()` instead. The implementation is automatically created by Gradle.

Also, annotated it with `@Input`.

melix (changes requested):
Thanks for the PR!

melix:
the test logs show `Mark property 'javaLauncher' as optional`, which I think is correct, it should be `@Optional`

PR #1151, opened 2025-08-16
Fix for kotlin 2

altro3:
Not micronaut gradle plugin is incompatible with `kapt` plugin with version `2.2.10` because DSL was changed and now you can't use KaptArguments with non-string values.

KAPT 1.9.25:
{624F2AED-FDDF-4EBC-815E-E0685E6A4B7B}

KAPT 2.2.10:

{E255C809-6D83-4290-81FA-9743887B2D1A}

This is crittical error, because now you can't use micronaut gradle plugin with gradle 9.0 + kotlin 2 + kapt


altro3:
@melix @graemerocher I think, it's very important fix

Issue #1165, opened 2025-09-30
Updating to 4.5.5 from 4.5.5 causes naligned versions of the junit-platform-engine a


sdelamo:
### Expected Behavior

```
Caused by: org.junit.platform.commons.JUnitException: OutputDirectoryProvider not available; probably due to unaligned versions of the junit-platform-engine and junit-platform-launcher jars on the classpath/module path.
```

A workaround is to add `org.junit.platform:junit-platform-launcher` as a `testRuntimeOnly` dependency.

melix:
I think this is a side effect of a leakage of junit-platform-launcher in previous releases of NBT. Adding `junit-platform-launcher` as a dependency is expected, that's how Gradle does it now, but we can probably do it in our plugin by convention.

sdelamo:
I think we should. This is going to be a puzzled breaking change for users.

PR #1166, opened 2025-10-03
Automatically add `junit-platform-launcher`

melix:
Older Gradle releases used to add this dependency automatically. Since Gradle 8, this is no longer the case (see https://docs.gradle.org/8.12/userguide/upgrading_version_8.html#manually_declaring_dependencies).

However, Native Build Tools used to leak this dependency, which caused other issues like duplicate entries on classpath. Since we upgraded to 0.11.1, this is no longer the case.

Therefore, users who upgrade may see an error if they don't add this dependency explicitly. To workaround this and make transition smoother, we're adding the dependency back implicitly.


graemerocher (approved):

melix:
Closes #1165

PR #1203, opened 2025-12-17
Detect IntelliJ Sync

melix:
This commit detects if IntelliJ is synchronizing the project classpath, in which case we don't add any test resources dependency, nor enable test resources, so that these are not added to the libraries that the user would see in the IDE.


dstepanov (approved):

melix:
I don't think there's anything official, but it works.

PR #1258, opened 2026-04-10
Fix applicationDefaultJvmArgs support

### LINKED ISSUE #765: Support applicationDefaultJvmArgs
### Feature description

The plugin does not support `application.applicationDefaultJvmArgs` 

I'm using `id("io.micronaut.application") version "3.7.10"` with gradle 8.2

mikehearn:
Fixes #765

### DIFF
--- a/minimal-plugin/src/main/java/io/micronaut/gradle/ApplicationPluginUtils.java
+++ b/minimal-plugin/src/main/java/io/micronaut/gradle/ApplicationPluginUtils.java
@@ -28,6 +28,10 @@ public final class ApplicationPluginUtils {
 
     public static List<String> applicationDefaultJvmArgs(Project project) {
         JavaApplication javaApplication = project.getExtensions().findByType(JavaApplication.class);
+        return applicationDefaultJvmArgs(javaApplication);
+    }
+
+    private static List<String> applicationDefaultJvmArgs(JavaApplication javaApplication) {
         if (javaApplication == null) {
             return List.of();
         }
@@ -39,6 +43,7 @@ public final class ApplicationPluginUtils {
     }
 
     public static Provider<List<String>> applicationDefaultJvmArgsProvider(Project project) {
-        return project.provider(() -> applicationDefaultJvmArgs(project));
+        JavaApplication javaApplication = project.getExtensions().findByType(JavaApplication.class);
+        return project.provider(() -> applicationDefaultJvmArgs(javaApplication));
     }
 }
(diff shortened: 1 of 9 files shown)


melix (changes requested):
One reason I didn't investigate this yet is the meaning of "default" JVM args. To me default means that if you have custom args, then the defaults should be ignored. As far as I understand in your PR, the arguments are merged. I'm not saying I'm against this, but minimally this should be listed as a potential breaking change, since applications which didn't have arguments can suddenly have arguments coming by upgrading, that's the reason for requesting changes (release notes).

mikehearn:
Where are breaking changes noted? It's targeted to master which is at 5.0.0-SNAPSHOT, so it should already be only in the next breaking release.

The problem with the current design is that from the user's perspective, they set `applicationDefaultArgs` which is supposed to be the default _for their project_ and then the args are silently eaten by the Micronaut plugin. This is confusing and really shouldn't be an expected behavior. I would expect that if I set the defaults then that's what I get unless I override them myself elsewhere. Micronaut just wipes out my work, currently.

melix:
Right but my comment is not about the fact you _use_ the default args, it's on the _semantics_ of default. If there _are_ arguments, my thinking is that you should not merge, and that default args should only be used if no arguments were provided.

mikehearn:
@melix I see. How does this look?

minimal-plugin/src/main/java/io/micronaut/gradle/ApplicationPluginUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+        var jvmArgs = new ArrayList<String>();
+        for (Object jvmArg : javaApplication.getApplicationDefaultJvmArgs()) {
+            jvmArgs.add(String.valueOf(jvmArg));
+        }
+        return jvmArgs;
+    }
+
+    public static Provider<List<String>> applicationDefaultJvmArgsProvider(Project project) {
+        return project.provider(() -> applicationDefaultJvmArgs(project));
```
melix: :x: Did you try this with the configuration cache? I think since you're using the project in the provider itself, this is going to fail.

mikehearn: Does Micronaut actually fully support the configuration cache? I must admit, I've never been able to get it to work for any project beyond a trivial hello world. Or is the support new in Micronaut 5/Gradle 9?

melix: We support the config cache, yes. There are users who report issues from time to time, so we must make sure to do our best diligence in being compatible.

mikehearn:
@melix OK, I think it's fixed now. I investigated why the tests didn't fail for me locally when it was cache-incompatible. It seemed to be something like, the relevant test can fail but still report true, if the failure is due to some environment thing like not having Docker running.

Anyway, it seems to pass CI now.

alvarosanchez:
@melix please re-review

mikehearn:
@melix Could you approve / merge please?

PR #1261, opened 2026-04-16
docs: Clarify packaging options and document the runner JAR

### LINKED ISSUE #1213: Docker runner JAR output locations doesn't mesh with Docs
### Expected Behavior

When running `./gradlew assemble`,  I expect the `build/libs` directory to contain only my application JAR, and for the `build/docker/main/layers` directory to be "runnable", as per [the docs](https://micronaut-projects.github.io/micronaut-gradle-plugin/latest/#_layered_application).

### Actual Behaviour

When the `runnerJar` task executes as part of the Docker plugin, it produces the JAR in the `build/libs` directory.

This isn't a bug in the "it doesn't work" sense. Everything mostly works just fine. However,

1. This can be confusing as, depending on which other plugins are available, a bunch of other JARs might end up in that directory as well.
2. The runner JAR can't do anything on its own. It's really more of an intermediate artifact for most workflows.
3. The [docs indicate](https://micronaut-projects.github.io/micronaut-gradle-plugin/latest/#_layered_application) the `build/docker/main/layers` directory should be "runnable". Presently, no JAR gets produced into that directory directly. However, if the runner JAR were placed there, then things would indeed work mostly as documented.

I actually began modifying the docs to address this, but then realized that moving the runner JAR output might be the more desirable change instead. Something along the lines of

```java
jar.getDestinationDirectory().set(project.getLayout().getBuildDirectory().dir("docker/main/layers"));
```
though I haven't tested this or checked to see if it should use constants or other task properties.

Anyhow, I'm happy to make a PR with whichever change you feel is more appropriate: clarification in the docs, or changes to the plugin.

### Steps To Reproduce

1. Start a start project using the Micronaut Application Gradle plugin
2. `./gradlew assemble`

### Environment Information

MacOS, JDK 21

### Example Application

_No response_

### Version

id("io.micronaut.platform.catalog") version "4.6.1"

MariusVolkhart:
## Summary

Fixes #1213

- **Replace misleading "Layered application" section** with "Standard distribution" — the old section described `build/docker/main/layers` as a runnable directory, but that's an internal artifact of the Docker build layers, not a supported way to run the application. Points users to `build/distributions` instead.
- **Add "Runner JAR" note** explaining that the Docker plugin produces an extra JAR in `build/libs/` that is used internally by the Docker build layers and is otherwise irrelevant. This addresses the confusion described in #1213 about unexpected JARs appearing in `build/libs/`.
- **Fix Kotlin DSL syntax** for ShadowJar `zip64` (use idiomatic property assignment instead of `.set()`)

## Context

I considered moving the runner JAR's output directory out of `build/libs/` (as suggested in the issue), but that conflicts with `BuildLayersTask` which owns `build/docker/main/layers` as its `@OutputDirectory` and deletes it at the start of each run.

## Test plan

- [x] Documentation renders correctly in the generated site
- [x] Existing `BuildLayersSpec` and `DockerBuildTaskSpec` tests continue to pass (no code changes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/docs/asciidoc/index.adoc
+++ b/src/docs/asciidoc/index.adoc
@@ -1129,6 +1129,10 @@ tasks.withType(io.micronaut.gradle.docker.tasks.BuildLayersTask) {
 }
 ----
 
+==== Runner JAR
+
+The plugin produces an additional "runner" `Jar`. This is used by the Docker build layers, but is otherwise irrelevant
+
 
 === Micronaut Runtimes
 
@@ -1251,13 +1255,11 @@ dockerfileNative {
 === Packaging the application
 
 By default, the plugin doesn't create a runnable fatjar when running `./gradlew assemble`.
-There are a couple of options:
 
-==== Layered application
+==== Standard distribution
 
-The plugin creates a "layered" application in `build/docker/main/layers` and from that directory you can run `java -jar myapp.jar`.
-It works because that directory contains a `lib` directory with all the libraries and a `resources` directory with the configuration.
-Keep in mind that copying the only `.jar` file to another directory won't work.
+Use the standard Gradle application distributions provided in `build/distributions`. These come with scripts to run your
+application, and package all the dependencies independently.
 
 ==== Add Shadow plugin
 
@@ -1545,7 +1547,7 @@ tasks.withType(ShadowJar).configureEach {
 [source, kotlin, subs="verbatim,attributes", role="multi-language-sample"]
 ----
 tasks.withType<ShadowJar>().configureEach {
-    zip64.set(true)
+    isZip64 = true
 }
 ----
 ====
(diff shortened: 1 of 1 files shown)


melix:
The "Replace misleading "Layered application" section" bullet point makes no sense to me. It mixes 2 orthogonal concerns. One is about docker layers, the other is the standard distribution. These are unrelated. AI generated?

MariusVolkhart:
It was hand-written, but unclear nevertheless.

The existing docs indicate the layers could be run directly, which doesn't work as documented. They also talk about copying the jar. I read that and assumed the docs are trying to show people a way to run the application from a terminal. Perhaps locally, or perhaps copying to an application server or something.

I changed the docs to point at the standard distribution, as I believe that's the best option for achieving that use case. However, in doing so I completely dropped the explanation of the docker layers. Perhaps that's where the confusion stems from. I'm happy to add that back in as a short section, just to answer the question of, "are you optimizing docker layers?"

BTW @melix , this plugin is very cool! Thanks for all your work on this.

melix (approved):

melix:
Thanks, I approved the PR because indeed the docs are confusing. I think we'll also need to improve the docs for the layered docker application setup, because that's missing (but can be done after this PR).

PR #1286, opened 2026-04-24
Track referenced OpenAPI files for task invalidation

alvarosanchez:
## Summary

Fixes #1227

- track recursively referenced local OpenAPI files as Gradle task inputs for shared generator tasks
- resolve referenced files by parsing structured YAML and JSON so only real `$ref` keys are followed
- add regressions for recursive local refs, ref-like text in comments/examples, and self-referential YAML aliases

## Test plan

- [x] `./gradlew -p .paperclip/worktrees/DEV-106 :micronaut-openapi-plugin:test --tests 'io.micronaut.openapi.gradle.OpenApiClientGeneratorSpec.reruns client generation when a recursively referenced local file changes' --tests 'io.micronaut.openapi.gradle.OpenApiClientGeneratorSpec.ignores ref-like text in comments and example blocks for task inputs' --tests 'io.micronaut.gradle.openapi.tasks.OpenApiReferenceResolverSpec'`
- [x] `./gradlew -p .paperclip/worktrees/DEV-106 :micronaut-openapi-plugin:check`

## Release Board Context

- QA-selected Micronaut organization project: `5.0.0 Release`
- Ambiguity note: the public project set also contained `5.0.0-M3`, but `5.0.0 Release` remained the best fit for default-branch bug work on the unreleased `5.0.0` line.
- Live project state at PR creation time: the visible open/public Projects v2 list currently includes `5.0.0-M3` but does not include `5.0.0 Release`, so no project link could be applied in this run.

---
###### ✨ This message was AI-generated using gpt-5.4

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -16,6 +16,7 @@ mockserver = "5.15.0"
 log4j2 = "2.25.4"
 tomlj = "1.1.1"
 openrewrite = "7.32.2"
+swagger-parser = "2.1.38"
 
 micronaut-platform = "5.0.2" # This is the platform version, used in our tests
 micronaut-aot = "3.0.0"
@@ -51,6 +52,7 @@ micronaut-aot-std = { module = "io.micronaut.aot:micronaut-aot-std-optimizers", 
 micronaut-aot-cli = { module = "io.micronaut.aot:micronaut-aot-cli", version.ref = "micronaut-aot" }
 
 micronaut-openapi-generator = { module = "io.micronaut.openapi:micronaut-openapi-generator", version.ref = "micronaut-openapi" }
+swagger-parser-v3 = { module = "io.swagger.parser.v3:swagger-parser-v3", version.ref = "swagger-parser" }
 
 micronaut-testresources = { module = "io.micronaut.testresources:micronaut-test-resources-build-tools", version.ref = "micronaut-testresources" }
 
--- a/openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
+++ b/openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
@@ -1,0 +1,264 @@
+/*
+ * Copyright 2003-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.micronaut.gradle.openapi.tasks;
+
+import java.io.File;
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.net.URI;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+final class OpenApiReferenceResolver {
+
+    private static final Pattern URI_SCHEME_PATTERN = Pattern.compile("^[A-Za-z][A-Za-z0-9+.-]*:.*$");
+    private static final Pattern WINDOWS_ABSOLUTE_PATH_PATTERN = Pattern.compile("^[A-Za-z]:[\\\\/].*$");
+
+    private OpenApiReferenceResolver() {
+    }
+
+    static List<File> referencedFiles(File definitionFile, Iterable<File> parserClasspath) {
+        if (definitionFile == null) {
+            return List.of();
+        }
+        var discoveredFiles = new LinkedHashSet<File>();
+        try (var parser = OpenApiDocumentParser.create(parserClasspath)) {
+            collectReferencedFiles(normalize(definitionFile.toPath()), parser, discoveredFiles, new HashSet<>());
+        } catch (IOException ignored) {
+            return List.of();
+        }
+        return List.copyOf(discoveredFiles);
+    }
+
+    private static void collectReferencedFiles(Path definitionFile, OpenApiDocumentParser parser, Set<File> discoveredFiles, Set<Path> visitedFiles) {
+        if (!Files.isRegularFile(definitionFile) || !visitedFiles.add(definitionFile)) {
+            return;
+        }
+        String content;
+        try {
+            content = Files.readString(definitionFile, StandardCharsets.UTF_8);
+        } catch (IOException ignored) {
+            return;
+        }
+        Object document = parser.parseStructuredDocument(definitionFile, content);
+        if (document == null) {
+            return;
+        }
+        Path baseDirectory = definitionFile.getParent();
+        for (String reference : findReferences(document)) {
+            Path resolvedReference = resolveLocalReference(baseDirectory, reference);
+            if (resolvedReference == null || !Files.isRegularFile(resolvedReference)) {
+                continue;
+            }
+            discoveredFiles.add(resolvedReference.toFile());
+            collectReferencedFiles(resolvedReference, parser, discoveredFiles, visitedFiles);
+        }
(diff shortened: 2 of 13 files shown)


openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    private static Object parseStructuredDocument(String content) {
+        try {
+            return new Yaml().load(content);
+        } catch (YAMLException ignored) {
+            return null;
+        }
+    }
```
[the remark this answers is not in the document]
alvarosanchez: Addressed in `38bf70ca`.

The resolver now parses with `new Yaml(new SafeConstructor(new LoaderOptions()))`, so referenced-file discovery stays limited to simple YAML/JSON value types instead of general-purpose tagged construction. I also added `OpenApiReferenceResolverSpec.uses safe yaml loading for tagged values` to lock that behavior down.

---
###### ✨ This message was AI-generated using gpt-5.4

openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
```diff
@@ -0,0 +1,143 @@
...
+            }
+            try {
+                return normalize(Path.of(URI.create(referenceLocation)));
+            } catch (IllegalArgumentException ignored) {
+                return null;
+            }
+        }
+        Path effectiveBaseDirectory = baseDirectory == null ? Path.of(".") : baseDirectory;
+        return normalize(effectiveBaseDirectory.resolve(referenceLocation));
+    }
```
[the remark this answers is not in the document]
alvarosanchez: Addressed in `38bf70ca`.

I now special-case Windows drive-letter absolute paths before URI-scheme detection, so refs like `C:/schemas/request.json` and `C:\\schemas\\request.json` are treated as local filesystem paths instead of being dropped as non-`file:` URIs. The resolver spec now includes a focused regression for those forms.

---
###### ✨ This message was AI-generated using gpt-5.4

openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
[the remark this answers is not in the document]
alvarosanchez: I did not add a raw-text fallback here.

That tradeoff is intentional: this issue already went through one security round-trip because raw-text `$ref` scanning overreached into comments and example blocks, and the approved plan explicitly kept referenced-file discovery structure-aware and local-only. I addressed the concrete parser-surface concern by switching to SnakeYAML's safe constructor in `38bf70ca`; if we later find a real OpenAPI document that the generator accepts but this structured resolver rejects, the follow-up should stay narrowly structured rather than reintroducing text scanning.

---
###### ✨ This message was AI-generated using gpt-5.4

gradle/libs.versions.toml
```diff
@@ -16,6 +16,7 @@ mockserver = "5.15.0"
 log4j2 = "2.25.4"
 tomlj = "1.1.1"
 openrewrite = "7.32.2"
+snakeyaml = "2.4"
```
altro3: 2.6

openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
```diff
@@ -0,0 +1,153 @@
...
+                continue;
+            }
+            discoveredFiles.add(resolvedReference.toFile());
+            collectReferencedFiles(resolvedReference, discoveredFiles, visitedFiles);
+        }
+    }
+
+    private static Object parseStructuredDocument(String content) {
+        try {
+            return new Yaml(new SafeConstructor(new LoaderOptions())).load(content);
```
altro3: At the moment, the logic is not correct: some of the schemes can be described in json, and some in yaml - and this is normal, so we need to take into account whether we have json or yaml in front of us.

altro3 (changes requested):

gradle/libs.versions.toml
alvarosanchez: Committed `bc226915` to address this. SnakeYAML is now pinned to `2.6` in the version catalog.

---
###### ✨ This message was AI-generated using gpt-5.5

openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
alvarosanchez: Committed `bc226915` to address this.

The resolver now chooses the parser from the referenced file type: `.json` files are parsed with Jackson JSON, while YAML/YML/OpenAPI files continue through SnakeYAML safe loading. I also added resolver coverage for JSON refs into YAML files and for rejecting YAML-only syntax in `.json` files, so we do not accidentally treat JSON schemas as loose YAML.

---
###### ✨ This message was AI-generated using gpt-5.5

gradle/libs.versions.toml
```diff
@@ -16,6 +16,8 @@ mockserver = "5.15.0"
 log4j2 = "2.25.4"
 tomlj = "1.1.1"
 openrewrite = "7.32.2"
+jackson = "2.19.2"
```
altro3: @alvarosanchez You need to check the AI's correct operation—it often produces nonsense. In this case, version 2.19.2, which was released over a year ago.

If you're not tied to the Swagger model and are parsing as abstract objects, it's better to use the latest version of Jackson—3.x.

openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Set;
+import java.util.regex.Pattern;
+import org.yaml.snakeyaml.LoaderOptions;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
+import org.yaml.snakeyaml.error.YAMLException;
+
+final class OpenApiReferenceResolver {
+
+    private static final ObjectMapper JSON_MAPPER = new ObjectMapper();
```
altro3: I hope the standard mapper will be able to correctly parse any specification; perhaps a more correct solution would be to use the Swagger parser. But this is an assumption.

gradle/libs.versions.toml
alvarosanchez: Committed `fe69009d` to address this. The OpenAPI plugin no longer declares a direct Jackson version or `jackson-databind` catalog entry for referenced-file discovery. The resolver now uses the Swagger parser/core parsing stack already aligned with `micronaut-openapi-generator`, with an explicit `swagger-parser-v3:2.1.38` runtime dependency for the plugin.

---
###### ✨ This message was AI-generated using gpt-5.5

openapi-plugin/src/main/java/io/micronaut/gradle/openapi/tasks/OpenApiReferenceResolver.java
alvarosanchez: Committed `fe69009d` to address this. `OpenApiReferenceResolver` now parses structured OpenAPI input through Swagger parser utilities: JSON uses Swagger core's shared JSON mapper and YAML uses `DeserializationUtils.readYamlTree`, while preserving the existing local-only `$ref` traversal and the tests for JSON/YAML file-type behavior, aliases, and unsafe YAML tags.

---
###### ✨ This message was AI-generated using gpt-5.5

gradle/libs.versions.toml
```diff
@@ -16,6 +16,7 @@ mockserver = "5.15.0"
 log4j2 = "2.25.4"
 tomlj = "1.1.1"
 openrewrite = "7.32.2"
+swagger-parser = "2.1.38"
```
altro3, line 19: Here's what I think: this dependency exists in micronaut-openapi , but it's not managed. I suggest making it managed in the micronaut-openapi .bom file and simply using the version specified in micronaut-openapi here to avoid duplicate versions.

alvarosanchez: I do not think this PR should make `micronaut-openapi-generator` an `implementation` dependency just to inherit the parser version here.

The plugin currently keeps the generator on the user-configurable `openApiGeneratorClasspath`; users can override `micronaut.openapi.version`, and the worker then runs with that selected generator. If this plugin adds the generator itself to `implementation`, the plugin runtime starts carrying the build's default generator version regardless of the user's configured OpenAPI version, which is a broader behavior change than this task needs.

I agree the better long-term shape is for `micronaut-openapi` to manage `swagger-parser-v3` in its BOM/metadata so downstreams do not duplicate that version. Until that exists upstream, this PR keeps the explicit parser dependency local to the Gradle plugin runtime because the resolver now directly uses Swagger parser classes before the worker classpath is involved.

---
###### ✨ This message was AI-generated using gpt-5.5

altro3 (approved):

PR #1343, opened 2026-06-09
Fix test resources standalone shutdown lifecycle

### LINKED ISSUE #1342: Regression: test resources service no longer stays alive
### Expected Behavior

I don't know if the issue is in the Gradle plugin, or in Micronaut Test Resources itself, but there is a serious regression since Micronaut 5 (I think). I have used the latest 5.0.2 release, and what I'm seeing is that starting the test resources service in the background with `./gradlew startTRS` will no longer keep it alive when the application shuts down.

I noticed because my database content no longer survived cross-builds. It is a very serious UX bug IMO.

### Actual Behaviour

The test resources service is systematically shutdown when the app is shutdown, independently of whether it was started explicitly before or not.

### Steps To Reproduce

1. `mn create-app repro-tr-bug --features data-jdbc,postgres`
2. `cd repro-tr-bug`
3. start the test resources service in the background: `./gradlew startTRS` <--- this should stay alive cross Gradle sessions
4. start the application: `./gradlew run` -> we see the postgres DB being started
5. hit `CTRL+C` to interrupt the server -> server is shutdown, test resources service should stay alive
6. Run `./gradlew stopTRS` -> fails with `java.net.ConnectException: Connection refused` which is the symptom of a test resources service which has been shutdown

### Environment Information

_No response_

### Example Application

_No response_

### Version

5.0.2

alvarosanchez:
## Summary

- Consume each per-build Test Resources stop marker once before shutdown so stale Gradle daemon listeners cannot reuse an old `true` marker.
- Preserve ordinary build-owned service cleanup while allowing an explicitly started standalone service to survive later application builds until `stopTestResourcesService` runs.
- Add focused TestKit coverage for build-owned shutdown followed by standalone service reuse and explicit stop.

## Verification

- `git diff --check`
- `./gradlew :micronaut-test-resources-plugin:test --tests "io.micronaut.gradle.testresources.ApplicationTestResourcesPluginSpec.standalone test resources service survives after build-owned service shutdown" --rerun-tasks --no-daemon`

## Release Target

- Target branch: `master`.
- Release target: `5.0.x` patch line; repository source currently indicates `5.0.1-SNAPSHOT`.
- Micronaut organization project: `5.0.3 Release`. Ambiguity preserved from QA intake: `5.0.1` and `5.0.2` Platform boards are closed, while `5.0.3 Release` is the earliest open matching Micronaut Platform BOM release board.

## PR Assets

- Not applicable; this changes code and regression tests only, with no rendered output or generated artifact requiring upload.

Fixes #1342

---


melix:
I need an explanation of what introduced the regression. We cannot simply throw AI PRs and finger cross they don't break things. This used to work, so what broke it?

alvarosanchez:
Root cause traced to the Gradle plugin lifecycle, not to Test Resources core.

The plugin registers a `BuildSessionLifecycleListener` on Gradle's parent listener manager, and the internal start task writes a per-build stop marker for the service it starts. That marker is `true` for a build-owned service and `false` for standalone/shared mode. The listener used to read that marker on `beforeComplete()` and call `ServerUtils.stopServer(settingsDirectory)` when it saw `true`, but it never consumed the marker.

Because the listener is registered on the parent listener manager, a Gradle daemon can keep that listener after the build that created the marker has finished. That left this failure sequence possible:

1. An ordinary `run` or `test` build starts a build-owned Test Resources service and writes a `true` stop marker.
2. That build stops its service, but the marker remains on disk and the daemon-scoped listener remains registered.
3. A later `startTestResourcesService` starts an explicit standalone service and writes fresh server settings in the same project settings directory.
4. A subsequent `run` build connects to the existing standalone service and should not own shutdown.
5. The stale listener from the earlier build still sees the old `true` marker and calls `stopServer()` against the same settings directory, shutting down the standalone service.

So the bug was stale shutdown authority: a build-owned stop marker outlived the build that owned it. The user-visible regression shows up with `run` because `JavaExec` tasks are wired into the Test Resources lifecycle, so application runs participate in the same start/connect/shutdown path as tests.

The fix consumes the marker once: read it, delete it, then act on the captured value. That keeps the existing behavior for ordinary build-owned services, but prevents an old build's `true` marker from being reused to stop a later standalone service. The regression test now covers both sides: the first ordinary `run` service is stopped at build end, then the standalone service survives a later `run` until `stopTestResourcesService` stops it explicitly.

---
###### ✨ This message was AI-generated using openai/gpt-5.5

alvarosanchez:
Replacement PR opened after correcting commit authorship to the alvarosanchez GitHub identity: https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/1344

melix:
I don't understand the explanation. This does NOT explain what commit introduced the regression.

Issue #1347, opened 2026-06-12
Add Support For Watching Additional File Directories During Continuous Builds


RooHoo33:
### Feature description

I have a Micronaut project that splits some source code files up between seperate Gradle sourcesets and want Micronaut to automatically rebuild if I edit files in the source sets. Currently, Micronaut only rebuilds the application if there are changes to files in the main source set and will ignore any changes to files in other source sets. I have an [example project I created](https://github.com/RooHoo33/watch-extra-files-poc/tree/main) with a Kotlin class located in a separate `extras` source set. I would love to have Gradle automatically rebuild my application when I make changes to the source file in the `extras/` source set.

I have thrown together a small feature change in [my fork with a possible solution](https://github.com/micronaut-projects/micronaut-gradle-plugin/compare/master...RooHoo33:micronaut-gradle-plugin:master) if this is a feature that can be supported and can open a PR! I wasnt sure where to expose this config and ended up just adding a new field to the `micronaut {...}` extension called `additionalFilesToWatch`. ([Ref in the example repo I created](https://github.com/RooHoo33/watch-extra-files-poc/blob/main/build.gradle.kts#L94))

RooHoo33:
Anyone have a chance to look at this and or is there anything I can provide? Thanks!

graemerocher:
the forked change doesn't look unreasonable, maybe submit a PR and we can review it

RooHoo33:
Sounds good, thank you! I opened https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/1361

PR #1361, opened 2026-06-26
Allow watching outside fields during continuous run

RooHoo33:
Micronaut automatically watches any files in the main source set when running `./gradlew run -t` but there is no support for watching files outside of the main source set. This pr adds a new field `additionalFilesToWatch` to the root `micronaut {...}` plugin extension (because I couldnt find a better place to put it but am open to suggestions/feedback). Any files passed to the new  field will be resolved just like files in the main source set. If those files modified, Micronaut will trigger a rebuild of the app.

I created an [example repo](https://github.com/RooHoo33/watch-extra-files-poc/tree/main) that has a separate source set with its [files being watched using this new feature](https://github.com/RooHoo33/watch-extra-files-poc/blob/main/build.gradle.kts#L94) and ran it with a locally published version of this plugin with my change.

I am also assuming if this is approved and merged, I will need to update [the plugin docs](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/src/docs/asciidoc/index.adoc) but I wasnt sure if that should be done in this pr or a separate one

Implements https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/1347

### DIFF
--- a/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
+++ b/minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
@@ -37,11 +37,9 @@ import org.gradle.api.tasks.SourceSetOutput;
 import org.gradle.api.tasks.TaskContainer;
 
 import java.io.File;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static io.micronaut.gradle.PluginsHelper.resolveRuntime;
 
@@ -106,17 +104,23 @@ public class MicronautMinimalApplicationPlugin implements Plugin<Project> {
             if (project.getGradle().getStartParameter().isContinuous() || Boolean.getBoolean(INTERNAL_CONTINUOUS_FLAG)) {
                 SourceSet sourceSet = sourceSets.findByName("main");
                 if (sourceSet != null) {
+                    MicronautExtension micronautExtension = project.getExtensions().findByType(MicronautExtension.class);
                     var sysProps = new LinkedHashMap<String, Object>();
                     sysProps.put("micronaut.io.watch.restart", true);
                     sysProps.put("micronaut.io.watch.enabled", true);
                     FileCollection sourceDirectories = sourceSet.getAllSource().getSourceDirectories();
+                    FileCollection additionalFilePathsToWatch = micronautExtension.getAdditionalFilesToWatch();
                     //noinspection Convert2Lambda
                     javaExec.doFirst(new Action<>() {
                         @Override
                         public void execute(Task workaroundEagerSystemProps) {
-                            String watchPaths = sourceDirectories
-                                    .getFiles()
-                                    .stream()
+                            Stream<File> fileSources = Stream.concat(
+                                    sourceDirectories
+                                            .getFiles().stream(),
+                                    additionalFilePathsToWatch.getFiles().stream()
+
+                            );
+                            String watchPaths = fileSources
                                     .map(File::getPath)
                                     .collect(Collectors.joining(","));
                             javaExec.systemProperty("micronaut.io.watch.paths", watchPaths);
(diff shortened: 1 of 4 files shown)


minimal-plugin/src/test/groovy/io/micronaut/gradle/MicronautMinimalApplicationPluginSpec.groovy
```diff
@@ -84,6 +84,54 @@ class MicronautMinimalApplicationPluginSpec extends AbstractGradleBuildSpec {
         task.outcome == TaskOutcome.SUCCESS
         watchLine.contains 'src/main/groovy'
     }
+    def "additional files are passed when configuring watch paths"() {
```
RooHoo33, line 87:
```suggestion
        @Issue("https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/1347")
    def "additional files are passed when configuring watch paths"() {
```
Not sure if this is needed or helpful?

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
```diff
@@ -37,11 +37,9 @@
 import org.gradle.api.tasks.TaskContainer;
 
 import java.io.File;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
graemerocher: if you can revert the usage of * imports. Thanks

graemerocher (changes requested):

graemerocher:
@melix other than the minor issue with the imports I don't see any issue with this do you? Maybe some documentation would be good.

minimal-plugin/src/main/java/io/micronaut/gradle/MicronautMinimalApplicationPlugin.java
RooHoo33: done!

RooHoo33:
> Maybe some documentation would be good.

I just need to [update this file](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/src/docs/asciidoc/index.adoc#micronaut-library-plugin) right? I can do that in this pr or open a separate one if that is easier for you all.

melix:
This sounds reasonable. AFAIU this is independent from Gradle's file watching. I'm not very familiar with how the system property is actually used within Micronaut to reload, though.

graemerocher:
> > Maybe some documentation would be good.
> 
> I just need to [update this file](https://github.com/micronaut-projects/micronaut-gradle-plugin/blob/master/src/docs/asciidoc/index.adoc#micronaut-library-plugin) right? I can do that in this pr or open a separate one if that is easier for you all.

Yes and in this PR is fine

RooHoo33:
> This sounds reasonable. AFAIU this is independent from Gradle's file watching. I'm not very familiar with how the system property is actually used within Micronaut to reload, though.

I believe this `micronaut.io.watch.paths` property is really only used to kill the micronaut server (specifically it uses the [FileWatchRestartListener](https://github.com/micronaut-projects/micronaut-core/blob/c4e938a52ee457ae53535a9064bdb18b7ecd8bab/context/src/main/java/io/micronaut/runtime/server/watch/event/FileWatchRestartListener.java#L39) which reacts to events from the [DefaultWatchThread](https://github.com/micronaut-projects/micronaut-core/blob/c4e938a52ee457ae53535a9064bdb18b7ecd8bab/context/src/main/java/io/micronaut/scheduling/io/watch/DefaultWatchThread.java)). Once micronaut kills itself, gradle sees the JavaExec task has ended and can rerun any changed tasks using its own file watching. Then it will restart the run task causing the micronaut server to reboot.

RooHoo33:
Added a little section about this in the docs file! I am seeing some build failures but they don't appear to be related to my change? They look like dependency changes in funspecs.

graemerocher (approved):

RooHoo33:
@graemerocher sorry to ping you again but the pr build is failing and it doesn’t look related to my changes? Do I need to merge main back into my branch or kick off the build again?

graemerocher:
yeah try that and see if it passes

RooHoo33:
Merged but the builds waiting. Im guessing I dont have permission to kick off a build?

RooHoo33:
@graemerocher do you need to approve the pr build?

graemerocher:
running CI for it here https://github.com/micronaut-projects/micronaut-gradle-plugin/pull/1374

RooHoo33:
Thank you so much!

